PR #11, opened 2024-11-09
train, manager, dashboard: show world size on dashboard, manual replica_id, convergence tweaks

d4l3k:
We allow setting a manual replica_id so the workers aren't continuously being recreated.

World size now shows on dashboard for convenience.

Verified loss converges

Test plan:

Lighthouse + 2 train_ddp

20241109_13h13m55s_grim


fegin:
The dashboard is pretty cool!

PR #14, opened 2024-11-22
process_group: register via public API

d4l3k:
This registers the ProcessGroups via the public APIs (`Backend.register_backend` and `dist.new_group`) rather than modifying internal state.

This makes things a bit messier but fits better into the usual PyTorch init logic. This requires a global PG to be initialized. A future PR could detect if a PG is initialized and register this as the global PG if necessary.

Weirdly -- with how this works it ends up returning the exact same ProcessGroup object from `dist.new_group`. I'm not quite sure how that actually works. The Backend wrapper seems to unwrap itself when casted to PG.

Test plan:

```
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -102,7 +102,7 @@ class ProcessGroup(BaseProcessGroup):
     def getBackendName(self) -> str:
         raise NotImplementedError("not implemented")
 
-    def register(self, name: str) -> None:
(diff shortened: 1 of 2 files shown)


fegin:
I like this approach even though it requires the global PG.

fegin (approved):

fegin (approved):

PR #16, opened 2024-11-22
Specifiy the devices when registering the backend to avoid warnings

fegin:
If we don't specify the devices, c10d will print out the warning when we create a new group.

pytest process_group_test.py


d4l3k (approved):
LGTM

PR #18, opened 2024-11-27
docs: add sphinx documentation and add missing documentation

d4l3k:
This adds sphinx documentation following the setup from https://github.com/pytorch/torchx.

This will build it in Github Actions CI and then deploy it.

Notably, this only builds the latest version of the documentation and not historical versions.

Test plan:

```
cd docs
pip install -r requirements.txt
make livehtml
make html
```

CI


d4l3k:
CI passing -- landing to main to test deploy

PR #44, opened 2024-12-16
manager: rename step to start_step + small shutdown fix

d4l3k:
This is a small change to Manager to:

1. make it clearer by renaming `step` to `start_step`
2. shutdown the background thread to prevent leaks

Test plan:

```
pytest
```


d4l3k:
CI passed locally, landing

PR #45, opened 2024-12-17
Add _test_pg helper

H-Huang:
Fix https://github.com/pytorch-labs/torchft/issues/32. Add a helper method `_test_pg` that runs multiple collectives. 

Potential follow ups:
- Update `ErrorSwallowingProcessGroupWrapper` and `ManangedProcessGroup` to support the other collectives
- additional consolidation of the tests. Can parametrize them based off the process group class?
- validate output values (maybe not necessary since this validation is done at pytorch collective level and these PGs are all wrapping those)
- move tests to a separate test folder

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,14 @@ lintrunner -a
 
 ### Tests
 
+We use `pytest` as our testing framework. To execute a specific test, use the following command:
+
+```sh
+pytest torchft/process_group_test.py -k test_device_mesh
+```
+
+To run the entire suite of tests:
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -67,6 +67,14 @@ lintrunner -a
 
 ### Tests
 
+We use `pytest` as our testing framework. To execute a specific test, use the following command:
```
d4l3k, line 70: We should document the Rust tests as well via `cargo test`

d4l3k (approved):
LGTM

Issue #39, opened 2024-12-13
LocalSGD support


d4l3k:
This is a tracking issue for adding LocalSGD support into torchft.

This should be fairly straightforward as we can use the Manager + quorum in an outer loop and then use an allreduce only periodically copy of the weights.

Something like:

```py
manager = Manager(...)
model = ...

while True:
    for step in range(local_steps):
        inputs, labels = next(dataloader_iter)
        optimizer.zero_grad()
        criterion(model(inputs), labels).backwards()
        optimizer.step()

    # update quorum and PGs (could overlap with the optimizer steps above)
    manager.step()

    # free gradient memory to make room for averaged weights
    optimizer.zero_grad(set_to_none=True)

    # copy the model weights and start the allreduce mean
    # we need a temporary copy to gracefully handle failures
    params = {}
    for name, param in model.named_parameters():
        copy = param.detach().clone()
        manager.allreduce_grad(copy)
        params[name] = copy

    # this will wait for all transfers to complete succesfully
    if manager.should_commit():
        for name, param in model.named_parameters():
            param.copy_(params[name])
            del params[name]
```

d4l3k:
One of the additional points here is on when we allow rejoining/recovering. Our current implementation is quite rigid but with LocalSGD we may want more control for when we detect failing workers as well as when we allow them to recover to avoid blocking.

https://pytorch.slack.com/archives/C083HHTCU06/p1734466793734549?thread_ts=1734049373.379299&cid=C083HHTCU06

> I think for flexibility we should change when we increment the step count
If we rename start_step to get_quorum and add a recover_allowed  field to it that would let us call this multiple times per step, with the first time allowing recovery and the second time not.
We would then move the step incrementing to either should_commit or to a new explicit step/commit method


![image](https://github.com/user-attachments/assets/b5f9a7cc-7730-46a8-bae8-f4150d645d8e)

d4l3k:
For the quorum we have a few options:

1. detect early failures and restart via heartbeats somehow
2. add a proper join/shrink-only support in quorum (adds complexity to quorum)
3. add support for multiple quorums per lighthouse -- that way we can do tick/tock behavior so workers will always join the primary quorum and the members in the secondary quorum are only healthy workers. This avoids complicating the quorum algorithm, though it will require some fallback logic if we get stuck in the secondary quorum

d4l3k:
[...]

PR #47, opened 2024-12-18
local_sgd: initial version of fault tolerant LocalSGD

d4l3k:
This implements an initial version of LocalSGD. #39

This has some limitations:

* workers can only join at the start of the sync_every period
* in async mode, the joining worker will discard the first ``sync_every`` steps as it's treated as a single Manager step
* if any errors/workers fail we will only detect it at the end of ``sync_every`` steps
* This uses a copy of the weights in memory in order to restore the model on commit failure to guarantee we always step exactly ``sync_every`` steps on each worker. This copy will be required for DiLoCo to compute the pseudo gradient.
* We don't do any bucketization for the allreduce operation

Test plan:

```
pyre
pytest
```

This will require more testing as I've only tested it with a mocked Manager and no e2e/integration tests

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -1,0 +1,184 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+LocalSGD
+=========
+
+This module implements a fault tolerant version of LocalSGD and related methods.
+"""
+
+from typing import Any, Dict, List, Mapping, Optional
+
+import torch
+from torch import nn, optim
+
+from torchft.manager import Manager
+
+
+class LocalSGD(nn.Module):
+    """
+    LocalSGD is a model wrapper similar to DistributedDataParallel that
+    implements the algorithm described in https://arxiv.org/pdf/1805.09767
+
+    This will synchronize the model parameters periodically in a fault tolerant
+    way using a torchft Manager. The allreduce on the parameters will happen
+    every sync_every steps after the optimizer.step call.
+
+    To implement safe and fault tolerant, this requires a backup copy of the
+    weights. By default these are stored in CPU memory. If any error occurs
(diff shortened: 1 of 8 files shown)


torchft/local_sgd.py
```diff
@@ -0,0 +1,176 @@
...
+        ``forward`` must be called before this function.
+        """
+        assert self._started_step, "forward must be called before step"
+        self._started_step = False
+
+        self._local_step += 1
+
+        if self._local_step >= self._sync_every:
+            self._local_step = 0
+            self._average()
```
H-Huang, line 157: in the function description it is saying we `allreduce` the weights, but the function implemented below allreduces the grads?

torchft/local_sgd.py
```diff
@@ -0,0 +1,176 @@
...
+        if self._local_step == 0:
+            self._manager.start_quorum()
+
+        self._started_step = True
+
+        return self._model.forward(*args, **kwargs)
+
+    def step(self) -> None:
+        """
+        This should be called after the optimizer step.
```
H-Huang, line 143: hmm, i wonder instead of having the user call `step`, we modify the init of `LocalSGD` to pass in the optimizer as well and then add an optimizer hook (https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.register_step_post_hook.html#torch.optim.Optimizer.register_step_post_hook) to automatically call this logic?

torchft/local_sgd.py
d4l3k: allreduce_grads is a bit misnamed, I'll rename it as we're not reducing gradients here

The behavior is just an allreduce though with some rescaling logic on top

torchft/local_sgd.py
d4l3k: TIL -- that looks pretty useful

I do think we can potentially move this into the forward wrapper as well but it makes things a bit harder to reason about

torchft/local_sgd.py
d4l3k: done

torchft/local_sgd.py
d4l3k: done

PR #53, opened 2024-12-20
chore: compile -> compile_protos

Jackmin801:
There's a warning about it. Goes away with this change
```
warning: use of deprecated method `tonic_build::Builder::compile`: renamed to `compile_protos()`
  --> build.rs:10:10
   |
10 |         .compile(&["proto/torchft.proto"], &["proto"])?;
   |          ^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: `torchft` (build script) generated 1 warning
```


d4l3k:
Yeah, I changed that because I hit some issue locally on my laptop but I think I have an old version cached for some reason

PR #59, opened 2025-01-08
manager: add per request timeouts

d4l3k:
This adds per request timeouts to all the ManagerClient methods. It also forwards the timeout from the manager to the lighthouse for quorum operations.

Test plan:

updated unit tests

```
pytest
cargo test
```

### DIFF
--- a/torchft/manager_integ_test.py
+++ b/torchft/manager_integ_test.py
@@ -1,10 +1,11 @@
 import logging
 import threading
+import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from contextlib import ExitStack
+from contextlib import ExitStack, contextmanager
 from dataclasses import dataclass, field
-from os import sync
-from typing import Callable, Dict, List, Protocol, Set, Tuple
+from datetime import timedelta
+from typing import Dict, Generator, List, Protocol, Set, Tuple
 from unittest import TestCase
 
 import torch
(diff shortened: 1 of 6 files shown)


Jackmin801 (approved):
Nice! lgtm. Maybe add python tests for this as well?

d4l3k:
Updated with some unit tests and an integration test + added timeout field to `should_commit`.

torchft/manager_test.py
```diff
@@ -483,3 +487,25 @@ def test_manager_numerics(self, client_mock: MagicMock) -> None:
...
+            1,  # max_step
+            1,  # max_rank
+            2,  # max_world_size
+            False,  # heal
+        )
+
+        manager.start_quorum(timeout=timedelta(seconds=12))
+        self.assertTrue(manager.should_commit(timeout=timedelta(seconds=23)))
+
+        self.assertEqual(client_mock().quorum.call_count, 1)
```
Jackmin801, line 511: Im not sure if this actually tests that the timeout is passed correctly. Seems to just test that the _clients methods are called at all. Maybe check that quorum was called with the expected timeout?

torchft/manager_integ_test.py
```diff
@@ -423,3 +423,46 @@ def test_local_sgd_recovery(self) -> None:
...
+                TimeoutError,
+                "status: Cancelled, message.*Timeout expired",
+            ):
+                manager.start_quorum(timeout=timedelta(seconds=0.01))
+
+            with self.assertRaisesRegex(
+                TimeoutError,
+                "status: Cancelled, message.*Timeout expired",
+            ):
+                manager.should_commit(timeout=timedelta(seconds=0.01))
```
Jackmin801, line 468: This doesnt really test that methods now have method specific timeouts though. They could still timeout on the default timeout and pass. Maybe set a different timeout in Manager init from the one used by method and assert the method one was used?

d4l3k: added an elapsed time check -- the default is 60s so easy to verify that we're shorter

torchft/manager_test.py
d4l3k: added a check on the kwargs

fegin (approved):

PR #60, opened 2025-01-08
process_group: add PG init timeouts + automatically assign manager port

d4l3k:
This does two things:

1) add PG init timeouts for ProcessGroupGloo and ProcessGroupBabyGloo/NCCL -- this does not add a timeout for ProcessGroupNCCL as it doesn't block on initialization so we can rely on our existing future timeouts
2) automatically assigns the manager port rather than requiring it to be manually set

Test plan:

```
```

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
     from torchft.process_group import ProcessGroup
 
 MANAGER_ADDR_KEY: str = "manager_addr"
-MANAGER_DEFAULT_PORT: int = int(os.environ.get("TORCHFT_MANAGER_PORT", 29511))
+MANAGER_PORT_ENV: str = "TORCHFT_MANAGER_PORT"
 REPLICA_ID_KEY: str = "replica_id"
 
 T = TypeVar("T")
@@ -74,6 +74,12 @@ class Manager:
     """
     Manager manages the full fault tolerant training loop.
 
+    This requires the that the TCPStore specified by the store_addr and
+    store_port or MASTER_ADDR and MASTER_PORT environment variables to be
+    started prior to creating this manager. If using a modern version of
+    torchelastic this will already be the case. Otherwise, it should be started
+    via torch.distributed.init_process_group prior to creating this manager.
+
     NOTE: when saving periodic checkpoints you must save and restore the
     Manager's state_dict as well to avoid synchronization issues.
     """
@@ -84,7 +90,6 @@ class Manager:
         load_state_dict: Callable[[T], None],
         state_dict: Callable[[], T],
         min_replica_size: int,
-        port: int = MANAGER_DEFAULT_PORT,
         use_async_quorum: bool = True,
         timeout: timedelta = timedelta(seconds=60),
         rank: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


torchft/manager.py
```diff
@@ -147,6 +147,13 @@ def _manager_state_dict() -> Dict[str, T]:
 
         if rank == 0:
             hostname = socket.gethostname()
+
+            if port is None:
+                port_str = os.environ.get(MANAGER_PORT_ENV)
+                if port_str is None:
+                    port = 0
+                else:
+                    port = int(port_str)
```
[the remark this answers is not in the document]
d4l3k: done, thanks for the suggestion!

torchft/process_group_test.py
```diff
@@ -122,6 +123,16 @@ def test_gloo(self) -> None:
...
+    def test_gloo_timeout(self) -> None:
+        store = TCPStore(
+            host_name="localhost", port=0, is_master=True, wait_for_workers=False
+        )
+
+        store_addr = f"localhost:{store.port}/prefix"
+        pg = ProcessGroupGloo(timeout=timedelta(seconds=0.01))
+        with self.assertRaisesRegex(RuntimeError, "timeout after 10ms"):
+            pg.configure(store_addr, 0, 2)
+
```
Jackmin801, line 135: Would an assert that the time taken to raise the timeout is now much smaller than the default be good here? Would make sure that future code does not break the ability for the wrapper to set the underlying PG creation timeout.

torchft/process_group_test.py
```diff
@@ -178,6 +193,17 @@ def test_baby_gloo(self) -> None:
...
+        store = TCPStore(
+            host_name="localhost", port=0, is_master=True, wait_for_workers=False
+        )
+
+        store_addr = f"localhost:{store.port}/prefix"
+
+        a = ProcessGroupBabyGloo(timeout=timedelta(seconds=0.01))
+        with self.assertRaisesRegex(TimeoutError, "timed out after 0.01 seconds"):
+            a.configure(store_addr, 0, 2)
+
```
Jackmin801, line 206: Same suggestion as the gloo one about testing that timeout was changed

torchft/process_group_test.py
```diff
@@ -178,6 +193,17 @@ def test_baby_gloo(self) -> None:
 
         torch.testing.assert_close(at, bt)
```
Jackmin801, line 194: We can assert an exact value here? Would be more a check that all reduce actually does all reduce which might not be the intention here

torchft/process_group.py
```diff
@@ -615,8 +645,14 @@ def _worker(
...
-            # pyre-fixme[20]: expects argument options
-            pg = cls.PG_CLASS(store, rank, world_size)
+            try:
+                # pyre-fixme[20]: expects argument options
+                pg = cls.PG_CLASS(store, rank, world_size)
+            except Exception as e:
+                logger.exception(f"got exception in worker: {e}")
+                tx.put(e)
+                return
+            tx.put(None)
```
Jackmin801, line 655: Since PGWrapper now uses _create_pg instead of PG_CLASS instantiation, should we make the code more consistent by having `_create_pg` here too?

torchft/process_group.py
```diff
@@ -602,6 +628,10 @@ def configure(self, store_addr: str, rank: int, world_size: int) -> None:
         )
         self._p.start()
 
+        # fetch the status of the PG init
+        # if an exception was returned _get will throw
+        assert _get(rx, self._timeout) is None
+
```
Jackmin801, line 634: If `_get` will throw the Exception, why do we need this assertion?

torchft/process_group_test.py
d4l3k: I don't think we need to compare to the default -- the message includes the time (in this case 10ms)

torchft/process_group_test.py
d4l3k: since we have the time in the message I think that's sufficient

torchft/process_group.py
d4l3k: just to make sure we consume the value -- doesn't really matter what the output is as long as it's consistent

torchft/process_group_test.py
d4l3k: done

torchft/process_group_test.py
Jackmin801: Ah ok I see. All good then

Jackmin801 (approved):
Nice! lgtm

fegin (approved):

PR #61, opened 2025-01-08
feat: expose lighthouse join timeout

Jackmin801:


src/lib.rs
```diff
@@ -225,15 +225,22 @@ struct Lighthouse {
...
         py.allow_threads(move || {
             let rt = Runtime::new()?;
 
             let lighthouse = rt
                 .block_on(lighthouse::Lighthouse::new(lighthouse::LighthouseOpt {
                     bind: bind,
                     min_replicas: min_replicas,
-                    join_timeout_ms: 100,
+                    join_timeout_ms: join_timeout_ms,
                     quorum_tick_ms: 100,
```
c-p-i-o, line 244: While here, do we also want to make `quorum_tick_ms` configurable?
Also, if possible, can we add a small unit test. 
Can be done in a separate PR or TODO for someone :-)

c-p-i-o (approved):
Change LGTM. Minor comment above - but it's not blocking.

src/lib.rs
Jackmin801: Thanks for the suggestion. Implemented :)
Not sure how to test quorum_tick though, seems like this would be quite flaky

PR #64, opened 2025-01-10
[lighthouse] detect unhealthy participants via heartbeats

d4l3k:
This makes the quorum only consider participants which have heathly heartbeats. 

This also exposes heartbeat settings to both the lighthouse (for how long we should consider a replica unhealthy) and in the manager for how frequently we should heartbeat.

I do feel like the quorum code is getting a bit messy -- though, with the joiners changes that will likely happen soon I'll be refactoring/removing the room state which should clean this up significantly.

Test plan:

pyre

```
pytest
cargo test
```

We don't currently have an e2e test where the heartbeats timeout

I'll also run a manual test to ensure we get the expected behavior


src/lighthouse.rs
```diff
@@ -90,55 +113,83 @@ fn quorum_changed(a: &Vec<QuorumMember>, b: &Vec<QuorumMember>) -> bool {
...
 // Checks whether the quorum is valid and an explanation for the state.
-fn quorum_valid(state: &RoomState, opt: &LighthouseOpt) -> (bool, String) {
-    let mut first_joined = Instant::now();
+fn quorum_valid(
+    now: Instant,
+    heartbeats: &HashMap<String, Instant>,
+    state: &RoomState,
+    opt: &LighthouseOpt,
+) -> (bool, String) {
+    let mut first_joined = now;
```
Jackmin801, line 122: Not a request for change but a clarification. This actually isnt needed right? first joined will be min(join time of participants waiting in room).

Or is it used for if theres no participants in the room? I would think min replica should be the one catching this.

Jackmin801 (approved):
Nice! lgtm. tests can be in a follow up PR so not blocking :)

src/lighthouse.rs
d4l3k: that's basically what this is computing -- this is just the initial value before we calculate the minimum

src/lighthouse.rs
```diff
@@ -90,55 +113,83 @@ fn quorum_changed(a: &Vec<QuorumMember>, b: &Vec<QuorumMember>) -> bool {
...
+
+    let healthy_participants: HashMap<String, QuorumMemberDetails> = state
+        .participants
+        .clone()
+        .into_iter()
+        .filter(|(replica_id, _details)| {
+            let last_heartbeat = heartbeats.get(replica_id);
+            if last_heartbeat.is_none() {
+                return false;
```
c-p-i-o, line 131: Add an optional message?

c-p-i-o (approved):

src/lighthouse.rs
d4l3k: We print the "heathly" replica count in the status message below, the dashboard should make this pretty obvious if no replicas have heartbeated

d4l3k:
I just realized that this controls quorum but still includes those workers in the finalized quorum, need to fix that and update the tests

d4l3k:
Also caught a bug where we won't increase the world size if a fast quorum is found -- now with fast quorum we will gracefully increase if there are available workers

src/lighthouse.rs
```diff
@@ -89,56 +112,94 @@ fn quorum_changed(a: &Vec<QuorumMember>, b: &Vec<QuorumMember>) -> bool {
...
-fn quorum_valid(state: &RoomState, opt: &LighthouseOpt) -> (bool, String) {
-    let mut first_joined = Instant::now();
+// Checks whether the quorum is valid, the new quorum and an explanation for the state.
+fn quorum_compute(
+    now: Instant,
+    heartbeats: &HashMap<String, Instant>,
+    state: &RoomState,
+    opt: &LighthouseOpt,
+) -> (Option<Vec<QuorumMember>>, String) {
+    let healthy_participants: HashMap<String, QuorumMemberDetails> = state
```
kennyyu, line 122: perhaps it might be worth refactoring calculating healthy participants to a separate helper? Then the logic for quorum could be a bit simpler:

health checker component -- maintains the state of all healthy participants

quorum:
- uses the health checker component to see which participants are alive and their states

src/lighthouse.rs
```diff
@@ -89,56 +112,94 @@ fn quorum_changed(a: &Vec<QuorumMember>, b: &Vec<QuorumMember>) -> bool {
...
-            return (is_fast_quorum, format!("Fast quorum found!"));
+            return (
+                Some(candidate_participants),
+                format!("Fast quorum found! {}", metadata),
+            );
         }
     }
 
-    if state.participants.len() < opt.min_replicas as usize {
+    if healthy_participants.len() < opt.min_replicas as usize {
```
kennyyu, line 170: Do we maintain a sequence number for each participant? For example, we could have a race where:

1) min members == 5
2) currently 4 have heartbeated
3) the fifth is just about to heartbeat, but then this code runs first and the fifth gets excluded
4) repeat over and over

If we have sequence numbers for each checkin, then if the fifth one is on a lower sequence number (but still healthy), we know we should keep waiting until the fifth one bumps the sequence number up to the expected sequence number.

d4l3k: We don't have any sequence numbers -- this logic is structured off of Raft heartbeats which assumes that the heartbeating is much faster than the timeout interval. Default settings is to heartbeat every 100ms and unhealthy timeout at 1s.

We do have waiting logic but that only considers healthy replicas

src/lighthouse.rs
d4l3k: That's a fair point, we could pull it out but we still need some of this info in the compute_quorum itself. In a follow up PR we added a "split brain" prevention so we won't create a quorum with less than half of the known healthy workers

We also use it for logging purposes

PR #68, opened 2025-01-13
process_group: wait for futher_thread join before creating new one

dwancn:
This modified the configure behavior in ProcessGroupBaby so that we wait for the futher thread to be terminated before we create new ones. This fixed a race condition where futher queue might be closed while futher thread was still trying to get item from it.

Test plan:
Added test cases for reconfiguring ProcessGroupBaby and included assertions to make sure internal state is set as expected.

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -609,8 +609,14 @@ class ProcessGroupBaby(ProcessGroup):
         if self._rx is not None:
             self._rx.close()
         if self._future_queue is not None:
+            # wait for the future thread to exit and then close the queue
             self._future_queue.put(_QUEUE_CLOSE)
-            assert self._future_queue is not None
+            assert self._future_thread is not None
+            self._future_thread.join(timeout=10.0)
+            # pyre-ignore[16]: optional value is checked above
+            if self._future_thread.is_alive():
+                raise RuntimeError("future thread did not exit")
+            # pyre-ignore[16]: optional value is checked above
             self._future_queue.close()
 
         ctx = mp.get_context("spawn")
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -210,6 +210,42 @@ class ProcessGroupTest(TestCase):
         with self.assertRaisesRegex(TimeoutError, "timed out after 0.01 seconds"):
             a.configure(store_addr, 0, 2)
 
+    def test_reconfigure_baby_process_group(self) -> None:
+        store = TCPStore(
+            host_name="localhost", port=0, is_master=True, wait_for_workers=False
+        )
+        store_addr = f"localhost:{store.port}/prefix"
+
+        a = ProcessGroupBabyGloo()
+        a.configure(store_addr, 0, 1)
+        future_thread_1 = a._future_thread
+        future_queue_1 = a._future_queue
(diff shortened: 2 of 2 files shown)


torchft/process_group.py
```diff
@@ -610,7 +610,10 @@ def configure(self, store_addr: str, rank: int, world_size: int) -> None:
             self._rx.close()
         if self._future_queue is not None:
             self._future_queue.put(_QUEUE_CLOSE)
-            assert self._future_queue is not None
+        # wait for the future thread to exit and then close the queue
+        if self._future_thread is not None:
```
d4l3k, line 614: these should all be conditioned under the same variable -- future_queue woudn't be set if thread isn't set and vice versa, can simplify the if statements here

torchft/process_group.py
```diff
@@ -610,7 +610,10 @@ def configure(self, store_addr: str, rank: int, world_size: int) -> None:
...
+            self._future_thread.join()
```
d4l3k, line 615: can we add a timeout to this so we don't block forever and instead throw an error message if we get into a weird state?

d4l3k (changes requested):
Would also be nice to add a PR description and test plan as well

Thanks for contributing!

torchft/process_group.py
dwancn: done

torchft/process_group.py
dwancn: done, thanks for suggestion

d4l3k (approved):
Changes look good but looks like lint is failing

dwancn:
> Changes look good but looks like lint is failing

thanks a lot. fixed the lint issue.

d4l3k:
@dwancn lint is still failing -- are you using `lintrunner`?

torchft/process_group_test.py
```diff
@@ -210,6 +210,42 @@ def test_baby_gloo_timeout(self) -> None:
...
+        a.configure(store_addr, 0, 1)
+        futher_thread_2 = a._future_thread
+        futher_queue_2 = a._future_queue
+        p_2 = a._p
+
+        self.assertNotEqual(futher_thread_1, futher_thread_2)
+        self.assertNotEqual(futher_queue_1, futher_queue_2)
+        self.assertNotEqual(p_1, p_2)
+
+        # pyre-ignore[16]: optional
```
d4l3k, line 235: FYI in the future the easier thing to do is to just assert that it's not None after you grab it above

i.e.
```
futher_thread_1 = ...
assert futher_thread_1 is not None

...
```

d4l3k (approved):

torchft/process_group_test.py
dwancn: Thanks for the suggestion! Updated the patch again with suggested changes. Fixed some typos as well.

dwancn: BTW, I found I was unable to run `lintrunner -a` directly as suggested by CONTRIBUTING.md. `pyre check` works for me.

d4l3k: @dwancn can you open an issue for that? Good to know if there's any issues with it.

Not great to having to have multiple back and forths just for some small lint issues and lintrunner usually solves that pretty easily

PR #69, opened 2025-01-13
feat: fix security warnings in torchft

c-p-i-o:
Summary:
Update jinja to address dependabot reported alerts.
1. Jinja has a sandbox breakout through malicious filename
2. Jinja has a sandbox breakout through indirect reference to format method.

Test Plan:
Test on github

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,5 +5,5 @@ matplotlib
 papermill
 ipykernel
 ipython_genutils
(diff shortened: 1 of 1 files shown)


d4l3k (approved):
LGTM

d4l3k:
The security issue here isn't a big deal since it's only during docs generation but thank you for getting rid of the alert :)

PR #70, opened 2025-01-13
lighthouse, manager: remove room support

d4l3k:
This removes previously added room support in #48

This didn't work out in practice as expected and will be instead replaced with "shrink only" quorum support which is a bit easier to reason about and doesn't require absurdly long timeouts.

Test plan:

```
pytest
cargo test
```

### DIFF
--- a/proto/torchft.proto
+++ b/proto/torchft.proto
@@ -50,11 +50,7 @@ message Quorum {
 }
 
 message LighthouseQuorumRequest {
-    // room_id is the specific quorum channel to use. All workers/replicas
-    // participating in the quorum must specify the same channel.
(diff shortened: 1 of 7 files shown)


H-Huang (approved):
> This didn't work out in practice as expected

Just curious, how come the quorom room did not work? Was it too slow?

d4l3k:
@H-Huang in practice this ended up being too fragile and could end up with two rooms flip-flopping and thus split brain training. Since for LocalSGD the step times can be 40 minutes, in order to prevent that behavior you either need the join timeout extremely long or you need to set min_replicas to always be >=50% of the replicas which isn't feasible in all cases

PR #71, opened 2025-01-14
lighthouse/quorum: avoid split brain and add shrink_only support

d4l3k:
This modifies the quorum behavior with a few major changes:

1. the quorum won't complete unless at least half of the heartbeating replicas join the quorum (this is to avoid split brain conditions) except in fast quorum cases when we're growing the quorum
  Rational: in scale up scenarios we could end up in split brain cases where min_replicas=2, but total replicas = 10 and we end up flip/flopping the quorum with effectively two separate jobs
2. the join_timeout now only applies if there's a heartbeating replica which hasn't joined the quorum 
  Rational: since heartbeating starts when manager starts (often before model init), there's not much use in waiting out the full time since we probably won't have any workers join and we have at least min_replicas
  (TODO: should we remove fast quorum as we can now complete quorum without waiting for the entire join timeout?)
3. added `shrink_only` support which will only allow workers to drop out but no new ones to join
  Rational: this is intended to support long step times such as needed in LocalSGD

Test plan:

Updated/added quorum test cases and added asserts on the explainations

```
cargo test
```

### DIFF
--- a/src/lighthouse.rs
+++ b/src/lighthouse.rs
@@ -6,6 +6,7 @@
 
 use core::net::SocketAddr;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::sync::Arc;
 use std::time::Duration;
 use std::time::{Instant, SystemTime};
@@ -115,19 +116,21 @@ fn quorum_compute(
     opt: &LighthouseOpt,
 ) -> (Option<Vec<QuorumMember>>, String) {
     let heartbeats = &state.heartbeats;
-    let healthy_participants: HashMap<String, QuorumMemberDetails> = state
-        .participants
-        .clone()
-        .into_iter()
-        .filter(|(replica_id, _details)| {
-            let last_heartbeat = heartbeats.get(replica_id);
-            if last_heartbeat.is_none() {
-                return false;
+    let healthy_replicas: HashSet<&String> = heartbeats
+        .iter()
+        .filter_map(|(replica_id, last_heartbeat)| {
+            if now.duration_since(*last_heartbeat) < Duration::from_millis(opt.heartbeat_timeout_ms)
+            {
+                return Some(replica_id);
             }
-
-            now.duration_since(*last_heartbeat.unwrap())
-                < Duration::from_millis(opt.heartbeat_timeout_ms)
+            None
         })
+        .collect();
+
+    let healthy_participants: HashMap<&String, &QuorumMemberDetails> = state
+        .participants
+        .iter()
+        .filter(|(replica_id, _details)| healthy_replicas.contains(replica_id))
         .collect();
 
     let mut candidate_participants: Vec<QuorumMember> = healthy_participants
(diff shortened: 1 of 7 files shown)


src/lighthouse.rs
```diff
@@ -163,30 +185,49 @@ fn quorum_compute(
...
+                "No quorum, only have {} participants, need min_replicas {} {}",
                 healthy_participants.len(),
                 opt.min_replicas,
                 metadata
             ),
         );
     }
 
+    // Avoid split brain by requiring at least half of the known alive workers.
+    if healthy_participants.len() <= healthy_replicas.len() / 2 {
```
H-Huang, line 202: Would you not get split brain if you have equal sizes of healthy workers? E.g. healthy_participants= 5, healthy_replicas= 10.

d4l3k: This check blocks the quorum so for the positive case we'd end up with

```rs
if 5 <= 10 / 2 {
    return (None, ...)
}
```

so we would make a quorum of minimum size 6

Jackmin801:
> should we remove fast quorum as we can now complete quorum without waiting for the entire join timeout?

I would think its ok. The fast quorum case is now covered by having no heartbeating replica that hasnt joined

src/lighthouse.rs
```diff
@@ -138,16 +141,35 @@ fn quorum_compute(
...
+        if shrink_only {
+            candidate_participants = candidate_participants
+                .into_iter()
+                .filter(|p| prev_replica_ids.contains(&p.replica_id))
+                .collect();
+        }
+
         // Fast quorum is when all previous participants are still in the quorum
         // and we have enough participants to form a quorum.
         let is_fast_quorum = prev_quorum
```
H-Huang, line 175: instead of prev_quorum, should this check `candidate_participants` are all in prev_replica_ids?

src/lighthouse.rs
```diff
@@ -138,16 +141,35 @@ fn quorum_compute(
     // Sort by replica ID to get a consistent ordering across runs.
     candidate_participants.sort_by_key(|p| p.replica_id.clone());
 
+    let shrink_only = healthy_participants
```
H-Huang, line 144: do we expect to have only some members have `shrink_only=True` and some `shrink_only=False`? I noticed we take `any` but should all the members be consistent?

H-Huang (approved):
i think it looks good, just had a few questions

src/lighthouse.rs
d4l3k: That is an expected case -- in LocalSGD we could have members join at any time and be in the wrong spot in the training loop. Members in the right location should already be in the quorum. In that case, the members not participating will get an exception and it's up to the caller to handle this correctly

src/lighthouse.rs
d4l3k: In fast quorum we do want to allow it to grow which is why we don't check that all candidate_participants are in prev_replica_ids. There could be new candidates

I may also just delete this given the changes to join timeout below

d4l3k:
Changes for a follow up PR:

* automatic retries for the not included in quorum case up to the specified timeout
* set per RPC timeouts using tokio::time::timeout_at
* enable TCP keep alives

PR #72, opened 2025-01-15
Dont return quorum if requester isnt involved

Jackmin801:


d4l3k (approved):
Code looks good @Jackmin801  were you planning on adding a unit test for this logic?

PR #73, opened 2025-01-15
overhaul timeouts for Lighthouse, Manager, checkpoint server

d4l3k:
This overhauls the timeouts for all network operations to allow for long quorum timeouts in a safer way.

Notable changes:

* removes timeout from the Endpoint/Channel/Client in favor of using keep alives + server based timeouts
* adds timeouts to CheckpointServer
* requires timeouts to be passed for all Rust operations
* adds a new `quorum_timeout` field to Manager py so you can have a much longer quorum timeout

Test plan:

```
pytest
cargo test
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,9 @@
 
 pub mod lighthouse;
 pub mod manager;
+mod net;
+mod retry;
+mod timeout;
 
 use core::time::Duration;
 use std::env;
@@ -46,6 +49,7 @@ impl Manager {
         store_addr: String,
         world_size: u64,
         heartbeat_interval: Duration,
+        connect_timeout: Duration,
     ) -> PyResult<Self> {
         py.allow_threads(move || {
             let runtime = Runtime::new()?;
@@ -58,6 +62,7 @@ impl Manager {
                     store_addr,
                     world_size,
                     heartbeat_interval,
+                    connect_timeout,
                 ))
                 .map_err(|e| PyRuntimeError::new_err(e.to_string()))?;
             let handle = runtime.spawn(manager.clone().run());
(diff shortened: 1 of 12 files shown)


src/net.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+pub async fn connect(addr: String, connect_timeout: Duration) -> Result<Channel> {
+    retry_backoff(
+        ExponentialBackoff {
+            initial_backoff: Duration::from_millis(100),
+            max_backoff: Duration::from_secs(10),
+            timeout: connect_timeout,
+            factor: 1.5,
+            jitter: Duration::from_millis(100),
```
Jackmin801, line 29: should jitter be random?

Jackmin801: oh this is max jitter

Jackmin801 (approved):
lgtm. `try_parse_grpc_timeout` still seems magic to me but all good if it works!

torchft/checkpointing.py
```diff
@@ -40,36 +41,50 @@ class CheckpointServer(Generic[T]):
...
+                            self.send_header("Content-type", "text/plain")
+                            self.end_headers()
+                            self.err(
+                                f"invalid checkpoint requested, serving {step} but got {self.path}"
+                            )
+                            return
+
+                        self.send_response(200)
+                        self.send_header(
+                            "Content-type", "tensor"
```
c-p-i-o, line 75: application/octet-stream?

c-p-i-o (approved):

src/lib.rs
```diff
@@ -84,36 +87,33 @@ impl Manager {
...
 #[pymethods]
 impl ManagerClient {
     #[new]
-    fn new(py: Python<'_>, addr: String, timeout: Duration) -> PyResult<Self> {
+    fn new(py: Python<'_>, addr: String) -> PyResult<Self> {
         py.allow_threads(move || {
             let runtime = Runtime::new()?;
             let client = runtime
-                .block_on(manager::manager_client_new(addr, timeout))
+                .block_on(manager::manager_client_new(addr, Duration::from_secs(60)))
```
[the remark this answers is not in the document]
d4l3k: added a `connect_timeout` setting to Manager+ManagerClient, really should have done this a while back :)

torchft/checkpointing.py
d4l3k: just wrapped this in a try catch but changed

src/net.rs
d4l3k: renamed for clarity

PR #75, opened 2025-01-16
use torchx for manual many replica (20+) tests

d4l3k:
This adds a torchx component that can be used to run many replica tests locally. This is intended to make it easier to validate E2E behavior since it's now 2 commands to run an e2e job with `train_ddp.py`.

TorchX is a job launcher and the component we create will launch one role per replica group with torchelastic managing the workers within each group. Elastic uses a port range `29600 + replica_id`.

It also includes some small tweaks to logging, static types and the Lighthouse dashboard for monitoring.

Test plan:

```
torchft_lighthouse --min_replicas 2 --join_timeout_ms 10000 &

torchx run -- --replicas 20
```
Increasing the join timeout is required to avoid split brain issues when some replicas are recovering. 

From experimentation, join_timeout_ms must be longer than the recovery time otherwise workers will never recover.

20250116_14h48m54s_grim

```
pyre
```

### DIFF
--- a/torchft/torchx.py
+++ b/torchft/torchx.py
@@ -1,0 +1,76 @@
+"""
+This is a file for TorchX components used for testing torchft.
+"""
+
+import os
+from typing import Dict, Optional
+
+import torchx.specs as specs
+
+
+def hsdp(
+    *script_args: str,
+    replicas: int = 2,
+    workers_per_replica: int = 1,
+    max_restarts: int = 10,
+    script: str = "train_ddp.py",
+    env: Optional[Dict[str, str]] = None,
+    image: str = "",
+    h: Optional[str] = None,
+    cpu: int = 2,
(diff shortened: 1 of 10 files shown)


torchft/torchx.py
```diff
@@ -0,0 +1,76 @@
...
+                env=env,
+                entrypoint="torchrun",
+                args=cmd,
+            )
+        )
+
+    return specs.AppDef(
+        name="torchft",
+        roles=roles,
+    )
```
H-Huang, line 76: Maybe include the commands / explanations in the description, e.g.
```
torchft_lighthouse --min_replicas 2 --join_timeout_ms 10000 &
torchx run -- --replicas 20
```
in `CONTRIBUTING.md` or `README.md`?

H-Huang (approved):
looks good!

PR #76, opened 2025-01-21
Add DiLoCo

H-Huang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #76

## API Usage

```python
# LocalSGD example
model = SimpleModel()
optimizer = optim.SGD(model.parameters())
manager = create_autospec(Manager)
with LocalSGD(manager, model, optimizer, sync_every=2):
    for inp, label in dataloader:
        loss = model(inp).mean()
        loss.backward()
        optimizer.step()
        
# DiLoCo example
model = SimpleModel()
inner_optimizer = torch.optim.AdamW(
    m.parameters(), lr=4e-4, weight_decay=0.1, betas=(0.9, 0.95)
)
outer_optimizer = torch.optim.SGD(
    m.parameters(), lr=0.7, momentum=0.9, nesterov=True
)
manager = create_autospec(Manager)
with DiLoCo(manager, model, inner_optimizer, outer_optimizer, sync_every=2):
    for inp, label in dataloader:
        loss = model(inp).mean()
        loss.backward()
        inner_optimizer.step()
        # outer_optimizer is actually used every 'sync_every' steps but this is hidden from the user

```

## Changes
- Updated `LocalSGD` to be a context manager rather than a `nn.Module` wrapper.
- Added DiLoCo. This is a subclass of LocalSGD since a lot of code is shared
- Added test to validate global models and outer optimizers and same at every step

discussion doc: https://docs.google.com/document/d/11c5JwQpSzilrDvK-vNsgQhpXAihbMn-hTRC8y3LiGqY/edit?tab=t.0#heading=h.izo4yi6jz4mk

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -3,25 +3,29 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
-
 """
 LocalSGD
 =========
-
 This module implements a fault tolerant version of LocalSGD and related methods.
 """
-
-from typing import Any, Dict, List, Mapping, Optional
+import logging
+from types import TracebackType
+from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, Type
 
 import torch
 from torch import nn, optim
+from torch.nn.parameter import Parameter
+from torch.optim.optimizer import Optimizer
+from torch.utils.hooks import RemovableHandle
 
 from torchft.manager import Manager
 
+logger: logging.Logger = logging.getLogger(__name__)
 
-class LocalSGD(nn.Module):
+
+class LocalSGD:
     """
-    LocalSGD is a model wrapper similar to DistributedDataParallel that
+    LocalSGD is a context manager that
     implements the algorithm described in https://arxiv.org/pdf/1805.09767
 
     This will synchronize the model parameters periodically in a fault tolerant
@@ -68,18 +72,14 @@ class LocalSGD(nn.Module):
             pin_memory: Whether to pin the memory used for the backup of the model parameters.
         """
         super().__init__()
-
         self._manager = manager
         self._model = model
+        self._local_optimizer = optimizer
         self._local_step = 0
-        self._started_step = False
         self._sync_every = sync_every
         assert sync_every >= 1, "sync_every must be greater than or equal to 1"
-
         device = backup_device or torch.device("cpu")
-
         self._backup_parameters: Dict[str, torch.Tensor] = {}
-
         for name, p in self._model.named_parameters():
             t = torch.empty(*tuple(p.shape), dtype=p.dtype, device=device)
             if (
@@ -90,86 +90,76 @@ class LocalSGD(nn.Module):
                 t = t.pin_memory()
             self._backup_parameters[name] = t
 
+        self._hooks: List[RemovableHandle] = []
         # Need to copy the parameters to the host to be safe if we are on the first step.
         self._save_parameters()
 
-        optimizer.register_step_post_hook(self._step_post_hook)
+    def __enter__(self) -> "LocalSGD":
+        # Add optimizer hook which increments the local step counter and syncs if necessary
+        self._hooks.append(
+            self._local_optimizer.register_step_post_hook(self._step_post_hook)
+        )
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> bool:
+        # Handle any cleanup or error handling here
+        if exc_type is not None:
+            # If an exception occurred, restore parameters
+            self._restore_parameters()
+        # Clean up hooks
+        for hook in self._hooks:
+            hook.remove()
+        self._hooks.clear()
+
+        return False  # Propagate exceptions
 
     def _save_parameters(self) -> None:
-        # TODO: consider running copy on a separate stream
-        for name, p in self._model.named_parameters():
-            self._backup_parameters[name].copy_(p.data, non_blocking=True)
+        with torch.no_grad():
+            # TODO: consider running copy on a separate stream
+            for name, p in self._model.named_parameters():
+                self._backup_parameters[name].copy_(p.data, non_blocking=True)
 
     def _restore_parameters(self) -> None:
-        # TODO: consider running copy on a separate stream
-        for name, p in self._model.named_parameters():
-            p.data.copy_(self._backup_parameters[name], non_blocking=True)
-
-    # pyre-fixme[14]: support state_dict args
-    def state_dict(self) -> Dict[str, object]:
-        """
-        state_dict returns the state_dict from the last time LocalSGD
-        synchronized and not the current weights.
-        """
-        state_dict = self._model.state_dict()
-        for name, p in self._backup_parameters.items():
-            assert name in state_dict
-            state_dict[name] = p
-        return state_dict
-
-    def load_state_dict(
-        self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False
-    ) -> None:
-        """
-        Loads the state dict to the model and the backup parameters.
-
-        This must be called while the model weights aren't being modified to
-        avoid corrupting the backup weights.
-        """
-        self._model.load_state_dict(state_dict, strict=strict, assign=assign)
-        self._save_parameters()
-
-    def forward(self, *args: object, **kwargs: object) -> object:
-        """
-        Run the model parameters.
-
-        This should be called before the optimizer step.
-
-        This will start the quorum and save the parameters if this is the first step.
-        """
-        if self._local_step == 0:
-            self._manager.start_quorum()
-
-        self._started_step = True
-
-        return self._model.forward(*args, **kwargs)
+        with torch.no_grad():
+            # TODO: consider running copy on a separate stream
+            for name, p in self._model.named_parameters():
+                p.data.copy_(self._backup_parameters[name], non_blocking=False)
 
     def _step_post_hook(
         self, _optim: optim.Optimizer, _args: List[object], _kwargs: Dict[str, object]
     ) -> None:
         """
         This hook is registered on the optimizer and is called after the optimizer step.
-
-        This will call the allreduce on the model weights every sync_every steps.
-        If any errors occur it will restore to the weights from the previous sync.
-
-        ``forward`` must be called before this function.
         """
-        assert self._started_step, "forward must be called before step"
-        self._started_step = False
-
         self._local_step += 1
-
         if self._local_step >= self._sync_every:
-            self._local_step = 0
-            self._average()
+            self.sync()
 
-            if self._manager.should_commit():
-                # save the parameters so we can restore from them later if necessary.
-                self._save_parameters()
-            else:
-                # commit failed, restore from the backup parameters
-                self._restore_parameters()
+    def sync(self) -> None:
+        """
+        Synchronizes and averages the model weights across the manager.
+        """
+        self._manager.start_quorum()
+        self._perform_sync()
+        self._local_step = 0
+
+    def _perform_sync(self) -> None:
+        """
+        Performs the synchronization of the model weights across the manager.
+        This method is intended to be overridden by subclasses to implement custom
+        synchronization logic.
+        """
+        self._average()
+        if self._manager.should_commit():
+            self._save_parameters()
+        else:
+            # commit failed, restore from the backup parameters
+            self._restore_parameters()
 
     def _average(self) -> None:
         # TODO: do we need to broadcast buffers like DDP does?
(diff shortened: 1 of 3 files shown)


torchft/local_sgd.py
```diff
@@ -93,7 +98,30 @@ def __init__(
         # Need to copy the parameters to the host to be safe if we are on the first step.
         self._save_parameters()
 
-        optimizer.register_step_post_hook(self._step_post_hook)
+    def __enter__(self):
+        # Add optimizer hook which increments the local step counter and syncs if necessary
+        self._opt_hook = self._local_optimizer.register_step_post_hook(
```
d4l3k, line 103: ahh, this is a nice change! Thanks for eliminating the wrapper :)

torchft/local_sgd.py
```diff
@@ -105,71 +133,53 @@ def _restore_parameters(self) -> None:
...
-        This should be called before the optimizer step.
-
-        This will start the quorum and save the parameters if this is the first step.
+        Start the quorum before each module forward.
         """
         if self._local_step == 0:
             self._manager.start_quorum()
 
-        self._started_step = True
+    # def should_sync(self) -> bool:
```
d4l3k, line 153: nit remove dead code

torchft/local_sgd.py
```diff
@@ -182,3 +192,63 @@ def _average(self) -> None:
...
+        super().__init__(
+            manager, model, inner_optimizer, sync_every, backup_device, pin_memory
+        )
+        self._outer_optimizer = outer_optimizer
+
+    def _model_sync(self) -> None:
+        """
+        ensure model has the same weights
+        """
+        pass
```
d4l3k, line 224: what's this for?

torchft/local_sgd.py
```diff
@@ -182,3 +192,63 @@ def _average(self) -> None:
...
+        """
+        Overrides the sync method to calculate the pseugradient, average them across the manager group, and
+        step using the outer optimizer.
+        """
+
+        # Set the .grad field of each parameter to its pseudogradient
+        for name, p in self._model.named_parameters():
+            assert name in self._backup_parameters
+            pseudogradient = p.data - self._backup_parameters[name]
+            p.grad = pseudogradient
```
d4l3k, line 236: This is a nice touch -- avoids the extra memory usage from the 3rd copy of the grads :)

torchft/local_sgd.py
```diff
@@ -182,3 +192,63 @@ def _average(self) -> None:
...
+        Average the gradients across the diloco group.
+        """
+        works = []
+        for p in self._model.parameters():
+            # Perform allreduce on the pseudogradients
+            work = self._manager.allreduce(p.grad)
+            works.append(work)
+        # Wait for all allreduce operations to complete
+        for work in works:
+            work.wait()
```
d4l3k, line 254: nit: Doesn't need to be part of this PR but we should pull the reduction code into a helper that we could share. We'll want to bucketize both of these as well as a followup

torchft/manager_integ_test.py
```diff
@@ -229,25 +229,104 @@ def state_dict() -> Dict[str, Dict[str, object]]:
...
+        )
+
+        def load_state_dict(state_dict: Dict[str, Dict[str, object]]) -> None:
+            m.load_state_dict(state_dict["model"])
+            inner_optimizer.load_state_dict(state_dict["inner_optim"])
+            outer_optimizer.load_state_dict(state_dict["outer_optim"])
+
+        def state_dict() -> Dict[str, Dict[str, object]]:
+            return {
+                "model": m.state_dict(),
```
d4l3k, line 280: One concern I have with this implementation is that we're transferring the model state_dict not the backed up parameters. If the local training is too fast we might end up with a corrupted copy of the weights in async mode.

There's two cases I'm concerned about:

### 1. async mode

1. trainers 0,1 call start_quorum and get a new quorum 
2. trainer 1 is behind so fetches the checkpoint from trainer 0
3. trainer 0 completes a step so it's weights have changed
4. trainer 1 finishes fetching the checkpoint and the weights are now inconsistent since it was actively being changed while fetching

We can solve this one by adding an assertion that the manager is in sync mode -- which is probably the behavior users want regardless.

### 2. should_commit==False

1. trainers 0,1 call start_quorum and get a new quorum 
2. trainer 1 is behind so fetches the checkpoint from trainer 0
3. should_commit() == False
4. trainer 0,1 fallback to the backup parameters
4. trainer 1 never saved the new checkpoint parameters and thus rolls back to untrained weights

d4l3k (changes requested):
Code is looking good :) There's a couple of edge cases I want to discuss more

torchft/local_sgd.py
H-Huang: removed!

torchft/local_sgd.py
H-Huang: yeah I definitely agree!

torchft/local_sgd.py
H-Huang: This isn't used, but it was a reminder for me to figure out how to make sure the model across all the replica groups are the same. For LocalSGD this doesn't matter as much since each group will have the same model after step 1.

For DiLoCo we need to ensure each replica starts with the same model weights otherwise each replica will be different since we are only exchanging pseudogradients. 

In DDP, we broadcast from rank 0 to the entire group, maybe we should do the same? If we don't do this explicitly for the user then maybe we at least should validate that the models have the same weights? What are your thoughts there?

torchft/manager_integ_test.py
H-Huang: Good points. Yeah this implementation hasn't really considered checkpointing very well. Your solution for number 1 makes sense. Have two follow up questions:

1. How should `CheckpointServer` interact with `LocalSGD` and `DiLoCo`, should we be saving the backup parameters to the checkpoint server after each sync?
2. For number 2, when do we have `should_commit == False`? Is this just for failures?

d4l3k: @H-Huang 

1. With the current design for LocalSGD/DiLoCo we always transfer the backup parameters (i.e. the weights right after the outer LocalSGD/DiLoCo averaging step). Effectively we treat the primary weights as the "intermediate state" equivalent to the gradients from normal training and we only commit them at the end of the N inner steps/1 outer step. With sync mode this is less of a concern since we will always heal before training so the backup and primary weights will always be the same
2. If any error occurs during the cross replica communication we'll end up with should_commit == False, this most commonly will happen if a worker in a different replica group crashes and then NCCL/Gloo times out/errors. Should_commit checks whether any error has happened on any of the ranks in a replica group. This is intended to prevent partial (say 1/10 ranks had an error) and corrupted (a rank had an error in communication and tensor is in a partially updated state) model updates

In terms of automatically backing things up, I think we may be able to use `register_load_state_dict_post_hook` to detect when a state_dict is loaded to the model and trigger a backup when that happens.

torchft/local_sgd.py
wconstab: one middle ground could be to add a public API function to the DiLoCo object so users can opt into the extra validation step, but not get it by default.  I'm not sure it is defensible to insist on doing a broadcast of parameters, given models have gotten a lot bigger since the time DDP was developed.

e.g. 

```
with DiLoCo(m ...) as d:
    d.validate_model_init()
    for train steps...
```

Otoh as I write this, it occurs to me that this is really not a diloco specific feature.  It might be even better to just provide some kind of distributed util function that users can pass a local model obj to and can perform some cross-rank validation of parameter initialization.

wconstab:
> Updated LocalSGD to be a context manager rather than a nn.Module wrapper. This required adding a pre_forward_hook to the model start the quorum

haven't read the code yet, but having `model` as an argument to DiLoCo obj made me think about how this would work with pipeline parallelism, where there are multiple model parts.  

What is the reason that we need start-quorum to be connected to the model via pre-forward?  doesn't it just need to happen once per inner step?  Again, with PP, there are more than one forward per inner-optimizer step.  So we need to at least ensure we aren't starting a quorum too often.

torchft/local_sgd.py
```diff
@@ -90,86 +90,88 @@ def __init__(
...
 
-    def forward(self, *args: object, **kwargs: object) -> object:
+    def _forward_step_pre_hook(self, _module: nn.Module, _args: List[object]) -> None:
         """
-        Run the model parameters.
-
-        This should be called before the optimizer step.
-
-        This will start the quorum and save the parameters if this is the first step.
+        Start the quorum before each module forward.
```
wconstab, line 149: do we need a new quorum on every module forward?

we don't sync after every module forward/backward, do we?  IIUC we sync every 'sync_every'.  So do we only need a new quorum if we just finished a sync step?

I wonder whether this would be clearer if we inverted things
```
diloco = DiLoCo(...)

for steps:
   # every time we enter/exit this context, it signals a quorum start.  not sure if that is helpful, or maybe i'm misunderstanding things.
   with diloco():
        model()...
        optimizer.step()
```

wconstab: i guess riffing on that, the helpers for DiLoCo and LocalSGD seem fine to me (and convenient) but I think even more important in a way is providing a good base class people can build other recipes on top of.  The key pieces seem to be 
- expressing how often to sync
- expressing the code to run inside the quorum
- expressing the code to run across quorums

```
quorum = QuorumBuilder(outer_optimizer, ...)

for outer_steps:
   for inner_steps:
      with quorum:
          model()...
          inner_optimizer.step()
   if quorum.should_commit:
      average()
      outer_optimizer.step()

```

I am probably misusing some terms and apis here, and i'm not sure having separate loops is desirable, but i am wondering if we have building blocks that let users write both diloco and local-sgd in about that many lines of code themselves?

If we do, then building the DiLoCo/LocalSGD helpers on top seems nice too

fegin:
@H-Huang just fyi, ghstack doesn't work with TorchFT. You won't be able to land it later.

wconstab:
Why doesn't ghstack work? I thought it didn't need special enablement.
 Did you try using cmdline 'ghstack land'? @fegin

torchft/local_sgd.py
H-Huang: I think the distributed utility for validation would be helpful. I got around this requirement in my integration test by initializing in the main process then passing that `state_dict()` to the threads running the replica groups. In a real environment this wouldn't be possible so it would make sense to load from a checkpoint. We dont have a method that loads from checkpoint in the manager which is also blocking so I can implement that in a follow up PR

torchft/local_sgd.py
H-Huang: These are good points! I don't actually think we need the quorum for each forward step so i removed the pre-forward hook. The only downside to not computing quorum each step is the delayed failure detection. This would be a problem if dropping some replica groups was a problem for convergence, but the DiLoCo paperssaw only a small drop in perplexity compared to a perfect communication setting.

I like the more general building block pieces you mentioned. Currently `Manager` acts as the component that manages the quorums and what is happening across replica groups so in that same vein i think it makes sense to enhance it to also run custom code. In doing so, I think we can achieve something similar to what you wrote above and gives users better flexiblity

torchft/manager_integ_test.py
```diff
@@ -431,6 +517,108 @@ def test_local_sgd_recovery(self) -> None:
...
+                if str_step in state_dicts[0]:
+                    # inner optimizer will be different, outer optimizer and model should be the same
+                    torch.testing.assert_close(
+                        state_dict["backup_params"],
+                        state_dicts[0][str_step]["backup_params"],
+                    )
+                    torch.testing.assert_close(
+                        state_dict["outer_optim"],
+                        state_dicts[0][str_step]["outer_optim"],
+                    )
```
d4l3k, line 618: should we be checking "model" params as well? or are they equivalent to backup_params at the end of the step?

d4l3k (approved):
LGTM

torchft/manager_integ_test.py
d4l3k: As discussed offline, after this PR lands we still have issue 2. We should follow this up with a PR to support shrink_only=True quorum so we never discard any computation and drop the rollback behavior to mimic how people actually use LocalSGD in practice

cc @Jackmin801

torchft/local_sgd.py
```diff
@@ -60,26 +64,22 @@ def __init__(
...
-            optimizer: The optimizer used by the model.
-            sync_every: How often to sync the model weights.
-            backup_device: The device to store the backup of the model parameters on. (default cpu)
-            pin_memory: Whether to pin the memory used for the backup of the model parameters.
+            manager (Manager): The manager to use.
+            model (nn.Module): The model to wrap.
+            optimizer (optim.Optimizer): The optimizer used by the model.
+            sync_every (int): How often to sync the model weights.
+            backup_device (Optional[torch.device]): The device to store the backup of the model parameters on. (default cpu)
+            pin_memory (bool): Whether to pin the memory used for the backup of the model parameters.
```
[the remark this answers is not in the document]
H-Huang: Oops, no reason, I removed it

torchft/local_sgd.py
```diff
@@ -90,95 +90,150 @@ def __init__(
...
+        step using the outer optimizer.
         """
-        assert self._started_step, "forward must be called before step"
-        self._started_step = False
 
-        self._local_step += 1
+        # Set the .grad field of each parameter to its pseudogradient
+        for name, p in self._model.named_parameters():
+            assert name in self._backup_parameters
+            pseudogradient = p.data - self._backup_parameters[name]
```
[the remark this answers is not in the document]
H-Huang: Oh yeah good point, I will change up the backup_parameters and the load_state_dict in a follow up PR

torchft/manager_integ_test.py
H-Huang: Correct, they will be equivalent at the end of the step

PR #78, opened 2025-01-23
lib: fix Already borrowed

d4l3k:
This fixes errors from Rust when calling the manager from multiple threads.

```
[rank1]: Traceback (most recent call last):
[rank1]:   File "/home/mreso/torchft/train_fsdp.py", line 176, in <module>
[rank1]:     main()
[rank1]:   File "/home/mreso/torchft/train_fsdp.py", line 169, in main
[rank1]:     optimizer.step()
[rank1]:   File "/home/mreso/torchft/torchft/optim.py", line 53, in step
[rank1]:     if self.manager.should_commit():
[rank1]:   File "/home/mreso/torchft/torchft/manager.py", line 556, in should_commit
[rank1]:     should_commit = self._client.should_commit(
[rank1]: RuntimeError: Already borrowed
```

We have the Manager receiver type set as `&mut self` when it really doesn't need to be. We instead do a lightweight clone of the client when making network calls.

See https://github.com/PyO3/pyo3/discussions/2465 for more details

Test plan:

```
pytest
cargo test
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,7 +109,7 @@ impl ManagerClient {
     }
 
     fn quorum(
-        &mut self,
+        &self,
         py: Python<'_>,
         rank: i64,
         step: i64,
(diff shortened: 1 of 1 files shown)


mreso (approved):
LGTM

PR #81, opened 2025-01-25
checkpointing: use CheckpointTransport abstraction

d4l3k:
This is a major refactoring of the live checkpointing code.

New behavior:

* Uses round robin recovery rather than all workers recovering from the primary (new logic in compute_quorum_result)

Key refactors:

* made a new generalized `CheckpointTransport` which `CheckpointServer` implements. This abstraction is designed to be symmetric for recovery strategies that require both sender and receiver to be aware of the communication. This will allow for recovering via ProcessGroups/NCCL which can be significantly faster.
* Made `ManagerClient.quorum` return a `QuorumResult` struct rather than a named tuple
* Renamed `checkpoint_address` to `checkpoint_metadata` as not all transports will be using an address
* Refactored recovery logic into `compute_quorum_result` method and added better unit tests for it
* Added checkpoint_metadata tests (there were no checkpoint_address tests previously)

Test plan:

```
pytest
cargo test
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,7 +27,7 @@ pub mod torchftpb {
 }
 
 use crate::torchftpb::manager_service_client::ManagerServiceClient;
-use crate::torchftpb::{CheckpointAddressRequest, ManagerQuorumRequest, ShouldCommitRequest};
+use crate::torchftpb::{CheckpointMetadataRequest, ManagerQuorumRequest, ShouldCommitRequest};
 use pyo3::prelude::*;
 
 #[pyclass]
@@ -113,15 +113,15 @@ impl ManagerClient {
         py: Python<'_>,
         rank: i64,
         step: i64,
-        checkpoint_server_addr: String,
+        checkpoint_metadata: String,
         shrink_only: bool,
         timeout: Duration,
-    ) -> Result<(i64, i64, i64, String, String, i64, Option<i64>, i64, bool), StatusError> {
+    ) -> Result<QuorumResult, StatusError> {
         py.allow_threads(move || {
             let mut request = tonic::Request::new(ManagerQuorumRequest {
                 rank: rank,
                 step: step,
-                checkpoint_server_addr: checkpoint_server_addr,
+                checkpoint_metadata: checkpoint_metadata,
                 shrink_only: shrink_only,
             });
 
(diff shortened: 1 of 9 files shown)


src/manager.rs
```diff
@@ -407,6 +354,128 @@ impl ManagerService for Arc<Manager> {
...
+        .filter_map(|(i, p)| {
+            if p.step != max_step || max_step == 0 && primary.replica_id != p.replica_id {
+                Some(i)
+            } else {
+                None
+            }
+        })
+        .collect();
+    let all_recovering_ranks_set = all_recovering_ranks.iter().collect::<HashSet<_>>();
+    let serving_ranks: Vec<usize> = participants
```
H-Huang, line 418: nit: I was initially confused by the name serving_ranks, maybe "stable_ranks" or "up_to_date_ranks" would also make sense?

torchft/manager.py
```diff
@@ -470,29 +472,52 @@ def _async_quorum(
...
-                f"healing required, fetching checkpoint server address from {address=} {max_step=}"
-            )
-            primary_client = ManagerClient(
-                address, connect_timeout=self._connect_timeout
-            )
-            checkpoint_server_address = primary_client.checkpoint_address(
-                self._rank, timeout=self._timeout
-            )
+        if allow_heal:
+            if quorum.recovering_ranks:
```
H-Huang, line 476: the `if recovering_ranks` condition is always true if `if heal` is true as well right?

src/manager.rs
```diff
@@ -407,6 +354,128 @@ impl ManagerService for Arc<Manager> {
...
+                "replica {} not participating in returned quorum",
+                replica_id
+            ))
+        })?;
+
+    let step = participants[replica_rank].step;
+
+    // Compute the details for workers at max step.
+    let max_step = participants.iter().map(|p| p.step).max().unwrap();
+    let max_participants: Vec<&QuorumMember> =
```
H-Huang, line 387: just checking, once we reach into compute_quorum_results will there always be a nonzero number of ranks at max_step in the quorum?

src/manager.rs
```diff
@@ -407,6 +354,128 @@ impl ManagerService for Arc<Manager> {
...
+            step,
+            max_step,
+            recover_rank.unwrap()
+        );
+    }
+
+    Ok(ManagerQuorumResponse {
+        quorum_id: quorum.quorum_id,
+        // address is used for looking up the checkpoint server address.
+        recover_manager_address: primary.address.clone(),
```
H-Huang, line 464: how/why do we use the `recover_manager_address` if the checkpoint loading is done round robin?

src/lib.rs
```diff
@@ -194,6 +196,41 @@ impl ManagerClient {
...
+#[pymethods]
+impl QuorumResult {
+    #[new]
+    fn new() -> Self {
+        Self {
+            quorum_id: 0,
+            replica_rank: 0,
+            replica_world_size: 1,
+            recover_manager_address: "".to_string(),
+            recover_rank: None,
```
H-Huang, line 223: nit: many rename `recover_rank` to `recovery_source_rank`? Or add comment that clarifies this represents the rank from which the current replica will recover

H-Huang (approved):
Overall looks good! Had some comments/questions but partly they are just to clarify my understanding of the changes

torchft/manager.py
```diff
@@ -104,6 +104,7 @@ def __init__(
         port: Optional[int] = None,
         hostname: str = socket.gethostname(),
         heartbeat_interval: timedelta = timedelta(milliseconds=100),
+        checkpoint_transport: Optional[CheckpointTransport[Dict[str, T]]] = None,
```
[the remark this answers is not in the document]
d4l3k: I'm not really anticipating users implementing one from scratch but I do want to provide two implementations here. One with CheckpointServer and another that's PG based

Advanced users can then use those by themselves or patch things like CheckpointServer to provide more robust implementations for their use cases

torchft/manager.py
d4l3k: They're flipped, this checks if anyone will be requesting a checkpoint from this node where as `heal` below is whether this node will be requesting a checkpoint

`heal == quroum.recover_rank is not None` which is distinct from `quorum.recovering_ranks`

src/manager.rs
d4l3k: yes, there's always at least 1 member at max_step since it's defined as the maximum step of participating members and we'll always have at least 1 member

src/manager.rs
d4l3k: done

src/lib.rs
d4l3k: renamed to `recover_src_rank` and `recover_dst_ranks`

PR #83, opened 2025-01-28
process_group/ManagedProcessGroup: ensure quorum and PG is configured

d4l3k:
Without this we will start running operations before the process group has been reconfigured/initialized. This can result in crashes, timeouts and deadlocking

When using Manager `allreduce` this already is the behavior so we just need to make sure we match the behavior for ManagedProcessGroup.

This was observed in https://github.com/pytorch/torchtitan/pull/806

Test plan:

```
pytest
lintrunner -a
```

Testing w/ https://github.com/pytorch/torchtitan/pull/806

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -19,7 +19,6 @@ runtime users need to take care to not assume a static rank or world size.
 import logging
 import queue
 import threading
-from abc import ABC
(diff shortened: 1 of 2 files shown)


fegin (approved):
Thanks for the fix!

PR #84, opened 2025-01-29
coordination: expose new low level torchft coordination API

d4l3k:
This exposes LighthouseServer, ManagerServer and ManagerClient so power users can directly call them to implement custom fault tolerance strategies.

It also renames some modules (`_torchft` for Rust lib) and methods to indicate which ones are private and subject to change.

This _does not_ expose a quorum algorithm currently. The plan is to add a new "simple quorum" interface that allows for custom data to be passed between members instead of the data required by the Manager.

We may also want to expose a pluggable quorum algorithm that can be passed into LighthouseServer to customize that as well.

Test plan:

```
cd docs && make livehtml
pytest
cargo test
```

localhost_8001_coordination html(Surface Pro 7)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,15 +30,30 @@ use crate::torchftpb::manager_service_client::ManagerServiceClient;
 use crate::torchftpb::{CheckpointMetadataRequest, ManagerQuorumRequest, ShouldCommitRequest};
 use pyo3::prelude::*;
 
+/// ManagerServer is a GRPC server for the manager service.
+/// There should be one manager server per replica group (typically running on
+/// the rank 0 host). The individual ranks within a replica group should use
+/// ManagerClient to communicate with the manager server and participate in
+/// quorum operations.
+///
+/// Args:
+///     replica_id (str): The ID of the replica group.
+///     lighthouse_addr (str): The HTTP address of the lighthouse server.
+///     hostname (str): The hostname of the manager server.
+///     bind (str): The HTTP address to bind the server to.
+///     store_addr (str): The HTTP address of the store server.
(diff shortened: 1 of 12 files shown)


d4l3k:
cc @b0noI

src/lib.rs
```diff
@@ -30,15 +30,30 @@ use crate::torchftpb::manager_service_client::ManagerServiceClient;
...
+///
+/// Args:
+///     replica_id (str): The ID of the replica group.
+///     lighthouse_addr (str): The HTTP address of the lighthouse server.
+///     hostname (str): The hostname of the manager server.
+///     bind (str): The HTTP address to bind the server to.
+///     store_addr (str): The HTTP address of the store server.
+///     world_size (int): The world size of the replica group.
+///     heartbeat_interval (timedelta): The interval at which heartbeats are sent.
+///     connect_timeout (timedelta): The timeout for connecting to the lighthouse server.
```
H-Huang, line 47: just curious, is it the standard in rust for the documentation to be above the `struct` definition rather than the `new` method since the parameters for the constructor aren't shown here.

H-Huang (approved):
Neat! looks good

src/lib.rs
d4l3k: These comments aren't really standard rust. This format is specifically so pyo3 will pull in the comments so they show up in the Python documentation

Added a screenshot to the test plan so you can see how it renders in sphinx

Sphinx doesn't render `__init__` separately by default so it all gets dumped into the class doc string

H-Huang: cool, thanks for the SS. Looks great!

PR #86, opened 2025-01-29
Fix ManagedDeviceMesh composability issues

fegin:
There are missing gaps of ManagedDeviceMesh to be actually used in TorchTitan. This PR fixes the gpas:

1. ManagedDeviceMesh is now able to be torch.save()/torch.load().
2. ManagedDeviceMesh will lie if there are zero replicated group participants. Size 0 DeviceMesh will cause confusion for training loops.
3. Corretly returns coordinates.


d4l3k (approved):
LGTM

PR #87, opened 2025-01-29
Change how TorchFT manages user_state_dict

fegin:
This PR closes some state_dict gaps when integrating with TorchTitan:
1. User state_dict() and load_state_dict() functions can be initialized lazily.
2. Change weights_only to False for torch.load as we may have to load some non-tensor states.

### DIFF
--- a/torchft/checkpointing.py
+++ b/torchft/checkpointing.py
@@ -172,7 +172,9 @@ class CheckpointServer(CheckpointTransport[T]):
             data = f.read()
 
         reader = io.BytesIO(data)
-        return torch.load(reader, weights_only=True)
+        # We have to set weights_only to True as there are some non-tensor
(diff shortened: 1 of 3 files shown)


d4l3k (approved):
LGTM

torchft/checkpointing.py
```diff
@@ -172,7 +172,9 @@ def load_from_address(cls, address: str, timeout: timedelta) -> T:
             data = f.read()
 
         reader = io.BytesIO(data)
-        return torch.load(reader, weights_only=True)
+        # We have to set weights_only to True as there are some non-tensor
```
c-p-i-o, line 175:
```suggestion
        # We have to set weights_only to False as there are some non-tensor
```

PR #89, opened 2025-01-29
ProcessGroupBaby: support full suite of PG tests

d4l3k:
This makes ProcessGroupBaby pass the PG test suite and fixes some glaring gaps in functionality.

* AllreduceOptions (and others) are now supported through a custom option wrapper
* implements allgather, broadcast
* BabyWork supports calling .wait() and .get_future() simultaneously
* BabyWork will clean up even if .wait isn't called

This should allow ProcessGroupBabyNCCL to be used with TorchTitan

Test plan:

```
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -19,8 +19,21 @@ runtime users need to take care to not assume a static rank or world size.
 import logging
 import queue
 import threading
+from dataclasses import dataclass
 from datetime import timedelta
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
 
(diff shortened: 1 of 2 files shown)


torchft/process_group.py
```diff
@@ -695,15 +723,18 @@ def _worker(
...
                     tx.put(next_op_id)
                     next_op_id += 1
                 elif cmd == "wait":
                     op_id: int = op[1]
                     work[op_id].wait()
-                    del work[op_id]
                     tx.put(op_id)
+                elif cmd == "del":
+                    op_id: int = op[1]
+                    del work[op_id]
```
fegin, line 738: Should we wait first before deleting it?

fegin (approved):
LGTM

torchft/process_group.py
d4l3k: I don't think we need to? In normal situations the user code will have waited for it, or they will have gotten a future. If they have a future we still have a reference and don't need it in this map anymore

The only other common case would be if an error occured and we're cleaning things up/shutting down -- in that case waiting likely won't succeed so no reason to wait

PR #90, opened 2025-01-30
CheckpointServer: start in disallowed state + tests

d4l3k:
This fixes a regression that was introduced in #81. 

We need to start checkpoint server in disallowed state so new requests coming in will block until the first time we call `send_checkpoint`. This prevents requesters erroring out on the first request since the server only has step `-1`.

This also added a timeout on acquiring the lock in the server thread (since we already timeout the socket) so don't want to get stuck there.

Test plan:

Added `disallow_checkpoint` call in torchtitan

Added unit tests for this behavior

```
pytest torchft/checkpointing_test.py
```

### DIFF
--- a/torchft/checkpointing.py
+++ b/torchft/checkpointing.py
@@ -17,9 +17,10 @@ import socket
 import threading
 import urllib.request
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from datetime import timedelta
 from http.server import BaseHTTPRequestHandler
-from typing import Generic, List, Optional, TypeVar
+from typing import Generator, Generic, List, Optional, TypeVar
 
 import torch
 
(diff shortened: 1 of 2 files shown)


torchft/checkpointing.py
```diff
@@ -117,7 +141,9 @@ def do_GET(self):
                     # validate socket timeout is actually set
                     assert self.connection.gettimeout() == self.timeout
```
H-Huang, line 143: Would it make sense to add an assert to check self._step == -1 (or maybe initialize step to None and check that) to fail quickly rather than to wait and timeout?

H-Huang (approved):
Looks good!

torchft/checkpointing.py
d4l3k: We actually are doing that which is the issue. Down below we check if self._step == the requested step

We end up with a race condition between send_checkpoint and the remote worker requesting the checkpoint which causes the requester to get an error due to mismatched step

This lock allows us to have some tolerance when requesting a checkpoint. We could busy loop/retry on caller side but the lock makes the code simpler and a bit more efficient

PR #82, opened 2025-01-27
[WIP][RFC] Required changes for integration with TorchTitan

fegin:
Summary:
We are not going to land this PR, this PR may be further divided into several PRs.

PR #91, opened 2025-01-30
ProcessGroupBabyNCCL: support multiple streams and use event on start

d4l3k:
This overhauls how we handle ProcessGroupBabyNCCL cuda events and streams. It also enables the PG test suite for BabyNCCL similar to #89 which enabled it for BabyGloo.

Key changes:
* BabyNCCL now is stream aware and will allocate one CUDA stream per `stream_id` in the parent process. This avoids issues when multiple streams are using NCCL and avoids any unintentional synchronization by using a single stream in the child process.
* BabyNCCL uses a CUDA event to synchronize the start of the NCCL operation. Previously BabyNCCL would immediately start running the NCCL operation even if the launching stream hadn't completed yet.
* Added `_pg_tests` for BabyNCCL
* Unified the codepath and got rid of _BabyNCCLWork/WORK_CLASS
* Detect when the subprocess has crashed and error out early
* Propagate `device_id` to the child process

Test plan:

```
```

test integration with torchtitan #82 with patch P1722192431

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -19,6 +19,7 @@ runtime users need to take care to not assume a static rank or world size.
 import logging
 import queue
 import threading
+from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass
 from datetime import timedelta
 from typing import (
@@ -26,10 +27,10 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Generator,
     List,
     Optional,
     Tuple,
-    Type,
     TypeVar,
     Union,
     cast,
@@ -58,9 +59,9 @@ from torch.distributed.distributed_c10d import (
     BroadcastOptions,
     ReduceOp,
     Work,
-    _world,
 )
 from torch.futures import Future
+from torch.utils._pytree import tree_any
 
 if TYPE_CHECKING:
     from torchft.manager import Manager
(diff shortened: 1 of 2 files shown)


torchft/process_group.py
```diff
@@ -586,29 +586,59 @@ def __init__(
...
-        assert op_id == self._op_id
-        assert isinstance(event, torch.cuda.Event)
+def _is_any_cuda(obj: object) -> bool:
+    """
+    Returns true if any of the tensors in the object are CUDA tensors.
 
-        # Wait on Event makes the stream wait but not the CPU thread.
-        event.wait()
+    Supports lists, tuples, dicts, and tensors.
+    """
```
fegin, line 619: This can be simplified with pytree?
```
from torch.utils._pytree import tree_any

return tree_any(lambda obj: isinstance(obj, torch.Tensor) and obj.is_cuda)
```

torchft/process_group.py
```diff
@@ -716,23 +750,76 @@ def _worker(
...
                 op = rx.get()
                 cmd = op[0]
                 if cmd == "func":
-                    func_name, args, kwargs = op[1:]
-                    args = _PickleSafeOptions.unsafe_args(args)
-                    fn = getattr(pg, func_name)
-                    work[next_op_id] = fn(*args, **kwargs)
+                    func_name, args, kwargs, stream_device, stream_id, event = op[1:]
+
+                    print(f"func {func_name=}")
```
fegin, line 763: Should we use logging?

torchft/process_group.py
```diff
@@ -716,23 +750,76 @@ def _worker(
...
+                        stream = streams[stream_key]
+                    else:
+                        stream = None
+
+                    with (
+                        torch.cuda.stream(stream)
+                        if stream is not None
+                        else nullcontext()
+                    ):
+                        print("stream created")
```
fegin, line 785: ditto

torchft/process_group.py
```diff
@@ -716,23 +750,76 @@ def _worker(
...
+                    # stream/synchronization behavior of the parent process.
+                    # We allocate one Stream per stream_id to make sure that we
+                    # don't accidentally introduce cross stream synchronization
+                    # points.
+                    if stream_id is not None:
+                        stream_key = f"{stream_device}/{stream_id}"
+                        if stream_key not in streams:
+                            streams[stream_key] = torch.cuda.Stream(
+                                device=stream_device
+                            )
```
fegin, line 775: Are we going to have zombie `stream` if there are multiple failures? Will this cause memory leakage?

torchft/process_group.py
d4l3k: Sorry, still need to clean these up

torchft/process_group.py
d4l3k: My understanding is streams are specific to the cuda context/process so this will be cleaned up just fine when it gets killed

torchft/process_group.py
d4l3k: done

torchft/process_group.py
d4l3k: removed

fegin (approved):

PR #98, opened 2025-02-05
[easy] README: fix logo font by using path instead of text

d4l3k:
This fixes the logo so it doesn't render as serif when on a device without the Freight Sans font.

Edited in inkscape with "Object to Path" to convert the FT from a `<text>` to a `<path>`.

Test plan:

Before:
Yi4U6OPyEe-1fwJCrBEABA

After:
fTc7tuPyEe-ZyQJCrBEAAg


H-Huang (approved):
nice!

PR #99, opened 2025-02-05
MonitoredQueue: fail fast when subprocess exits

d4l3k:
This is a wrapper around mp.Queue that fails fast when the subprocess exits. This means if subprocess BabyNCCL crashes we will report the error quickly rather than waiting for the timeout to elapse.

Test plan:

```
pytest -o log_cli=1 torchft/process_group_test.py torchft/multiprocessing_test.py -s -v -x
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -17,7 +17,6 @@ runtime users need to take care to not assume a static rank or world size.
 """
 
 import logging
-import queue
 import threading
 from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass
@@ -63,6 +62,8 @@ from torch.distributed.distributed_c10d import (
 from torch.futures import Future
 from torch.utils._pytree import tree_any
 
+from torchft.multiprocessing import _MonitoredQueue
+
 if TYPE_CHECKING:
     from torchft.manager import Manager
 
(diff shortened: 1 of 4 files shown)


torchft/process_group.py
```diff
@@ -573,8 +553,8 @@ class _BabyWork(Work):
     def __init__(
         self,
         pg: "ProcessGroupBaby",
-        tx: mp.Queue,
-        rx: mp.Queue,
+        tx: _MonitoredQueue,
```
H-Huang, line 556: What are `tx` and `rx` meant to represent?

torchft/process_group.py
```diff
@@ -739,7 +739,7 @@ def _worker(
             try:
                 pg = cls._create_pg(store, rank, world_size)
             except Exception as e:
-                logger.exception(f"got exception in worker: {e}")
+                print(f"got exception in worker: {e}", file=sys.stderr)
```
H-Huang, line 742: any particular reason for the logger -> print change?

H-Huang (approved):
LGTM!

torchft/process_group.py
d4l3k: We're not initializing the logger in the subprocess so switched to stderr to make sure we actually log these

Maybe I should just instantiate a second logger as well

torchft/process_group.py
d4l3k: These are the raw communication channels with the subprocess, might be good to refactor these so all queue comms go through the PG

d4l3k: refactored

Issue #97, opened 2025-02-04
process_group: support all PG APIs


d4l3k:
We currently only support a subset of collective APIs in `torchft.ProcessGroup`. We want to support all of them in all of our ProcessGroup implementations.

This can be done piece by piece and we don't need to add all of them right now.

The API definition for `ProcessGroup` is located at https://github.com/pytorch/torchft/blob/main/torchft/process_group.py#L123

We need to support these for at least ProcessGroupWrapper, ManagedProcessGroup, ProcessGroupBaby*.

We want to support all of the collectives in `PyProcessGroup`: https://github.com/pytorch/pytorch/blob/11f69808c64a65c68a4452250ba7719dcff27c78/torch/csrc/distributed/c10d/PyProcessGroup.hpp

* allgather_into_tensor_coalesced
* allreduce_coalesced
* alltoall_base
* barrier
* reduce_scatter
* reduce_scatter_tensor_coalesced
* send/recv (trickier)

Testing:

We have an existing test suite for collectives. We should add tests for these new types to `_test_pg` and make sure all of our PG implementations support them.

https://github.com/pytorch/torchft/blob/main/torchft/process_group_test.py#L96

PR #102, opened 2025-02-06
Adds `reduce_scatter` into `torchft`

allenwang28:
# What does this PR do?
Partially addresses https://github.com/pytorch/torchft/issues/97 by adding `reduce_scatter` into `torchft`.

Concretely, this consists of a few pieces: 
- Introducing `reduce_scatter` into the `ProcessGroup` following the signature [here](https://github.com/pytorch/pytorch/blob/11f69808c64a65c68a4452250ba7719dcff27c78/torch/csrc/distributed/c10d/PyProcessGroup.hpp#L203
- In `ProcessGroup*` we essentially follow the behavior of other collectives:
  - In `ProcessGroupWrapper`, it depends on the parent implementation
  - In `ProcessGroupDummy`, it writes from the first input into output
  - In `ProcessGroupBaby`, it asserts inputs and moves underlying storage into shared memory 
- Add `ReduceScatterOptions` in `_PickleSafeOptions`
- Introduces `reduce_scatter` as an option in `_test_pg`, however this necessitated a new function (named `_should_run_collective`) which was needed as e.g. GLOO does not support `reduce_scatter`. This function essentially takes the collective, backend and device and copies the logic of the [published supported collective matrix](https://pytorch.org/docs/stable/distributed.html#backends). 

# Tests
Presubmits, and:
```
$ pytest torchft/process_group_test.py 
============================================= test session starts =============================================
platform linux -- Python 3.10.16, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/allencwang/workspace/torchft
configfile: pytest.ini
plugins: typeguard-2.13.3
collected 16 items                                                                                            

torchft/process_group_test.py ................                                                          [100%]

============================================= 16 passed in 31.44s =============================================
[rank0]:[W206 14:54:24.777939032 CudaIPCTypes.cpp:16] Producer process has been terminated before all shared CUDA tensors released. See Note [Sharing CUDA tensors]
```

# Next steps
The logic of `_should_run_collective` is a bit confusing, as it allows "non defined backends" like `ErrorSwallowing*` through, to mimic the old behavior before this change. Testing here could become a bit unwieldy as we add more collectives and so a future step could be to refactor the testing. 

One nice change could be to parameterize tests by the collective. This will make potentially failing collectives more explicit and will reduce the time it takes to run individual tests. Likely can do this in the next PR.

### DIFF
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -61,6 +61,31 @@ def dummy_init_pg() -> None:
         )
 
 
+def _should_run_collective(collective_str: str, backend_str: str, device: str) -> bool:
+    """Verify if the collective is supported by the backend and device.
+
+    See https://pytorch.org/docs/stable/distributed.html#backends for the
+    supported collectives / backends / devices matrix.
+
+    """
+    if "nccl" in backend_str.lower():
+        # all collectives are supported for NCCL/CUDA but none on CPU.
+        return device == "cuda"
+    elif "gloo" in backend_str.lower():
+        if device == "cuda":
+            # GLOO/GPU only supports broadcast and all_reduce.
+            if collective_str in ["broadcast", "all_reduce"]:
+                return True
+            return False
+        else:  # cpu
+            if collective_str in ["reduce_scatter", "all_to_all"]:
+                return False
+            return True
+    else:
+        # Non defined backends (e.g. ErrorSwallowing) should continue to work.
+        return True
+
+
 def _test_pg(
     pg: ProcessGroup,
     example_tensor: torch.Tensor = torch.randn((2, 3), dtype=torch.float32),
@@ -95,10 +120,25 @@ def _test_pg(
         ("allgather", (output_tensors, [input_tensor], AllgatherOptions())),
         ("broadcast", (tensor_list, BroadcastOptions())),
         ("broadcast_one", (input_tensor, 0)),
-        ("reduce_scatter", (output_tensors, [input_tensor], ReduceScatterOptions())),
+        (
+            "reduce_scatter",
+            (output_tensors[0], [[input_tensor]], ReduceScatterOptions()),
+        ),
     ]
     works: Dict[str, dist._Work] = {}
+
+    try:
+        backend_str = pg.getBackendName()
+        device = example_tensor.device
+        if type(device) is torch.device:
+            device = device.type
+    except NotImplementedError as e:
+        backend_str = ""
+        device = ""
+
     for coll_str, args in collectives:
+        if not _should_run_collective(coll_str, backend_str=backend_str, device=device):
+            continue
         coll = getattr(pg, coll_str)
         work = coll(*args)
         works[coll_str] = work
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -317,7 +317,7 @@ class ProcessGroupTest(TestCase):
 
         store_addr: str = f"localhost:{store.port}/prefix"
 
-        def run(rank: int) -> Tuple[torch.Tensor, Work]:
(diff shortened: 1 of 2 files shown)


torchft/process_group_test.py
```diff
@@ -60,6 +61,31 @@ def dummy_init_pg() -> None:
...
+        # all collectives are supported for NCCL/CUDA but none on CPU.
+        return device == "cuda"
+    elif "gloo" in backend_str.lower():
+        if device == "cuda":
+            # GLOO/GPU only supports broadcast and all_reduce.
+            if collective_str in ["broadcast", "all_reduce"]:
+                return True
+            return False
+        else:  # cpu
+            if collective_str in ["reduce_scatter", "all_to_all"]:
```
d4l3k, line 81: oh wow -- didn't realize we don't support these on Gloo, good to know! cc @c-p-i-o

torchft/process_group_test.py
```diff
@@ -94,9 +120,25 @@ def check_tensors(arg: Any) -> None:  # pyre-ignore[2]
...
+        ),
     ]
     works: Dict[str, dist._Work] = {}
+
+    try:
+        backend_str = pg.getBackendName()
+        device = example_tensor.device
+        if type(device) is torch.device:
+            device = device.type
+    except NotImplementedError as e:
```
d4l3k, line 135: why are we getting a NotImplementedError? which backend is this?

torchft/process_group_test.py
fegin: ye, we miss many APIs on Gloo.

torchft/process_group_test.py
allenwang28: This is just for the `ErrorSwallowingProcessGroupWrapper`. I have a follow up PR to refactor the tests to get rid of this entire function though!

torchft/process_group_test.py
wconstab: this approach seems nice and explicit.  but is it possible to instead just try: the test, and except: some specific NYI error?  (i'm not sure if we raise a consistent type of NYI exception from backends?)

allenwang28:
Updated the test to be simpler, so I removed the utility functions I previously added. This should remove the need for a test refactor. I wanted to parameterize by collective, but https://github.com/pytorch/torchft/pull/103 shows that tests got much slower after doing this. I will deprecate #103.

I have also added an explicit `NotImplementedError` for `reduce_scatter` within `ProcessGroupBabyGloo`, because otherwise the exception lives within the `tx` queue (see [here](https://github.com/pytorch/torchft/blob/9533676281bf68f95eba6301b0b8eaef76bb9bec/torchft/process_group.py#L899-L910)). Noticed this as `test_baby_gloo_apis` would fail [here](https://github.com/pytorch/torchft/blob/9533676281bf68f95eba6301b0b8eaef76bb9bec/torchft/process_group_test.py#L267), and that would be a nasty issue for a downstream user. Therefore, adding this explicitly in the API is ultimately cleaner.

torchft/process_group_test.py
allenwang28: > this approach seems nice and explicit. but is it possible to instead just try: the test, and except: some specific NYI error? (i'm not sure if we raise a consistent type of NYI exception from backends?)

Yeah this is a good idea, I modified the block as follows:
```
    for coll_str, args in collectives:
        try:
            coll = getattr(pg, coll_str)
            work = coll(*args)
            works[coll_str] = work
            work.wait()
            fut = work.get_future()
            fut.wait()
            # Check that all tensor arguments have the expected shapes and dtypes
            check_tensors(args)
        except RuntimeError as e:
            if f"does not support {coll_str}" in str(e):
                # Skip collectives that are not supported by the backend.
                continue
            raise e
```

d4l3k (approved):
LGTM

wconstab (approved):

PR #103, opened 2025-02-07
Refactors `process_group_tests.py`

allenwang28:
# What does this PR do?
As part of https://github.com/pytorch/torchft/issues/97, this PR refactors `process_group_test`:
- Renames up `_test_pg` to `run_collectives` and extending it to accept a given list of collectives by name.
- Breaks up `ProcessGroupTest` into three tests: `GlooTest`, `NCCLTests` and `DummyTests`:
  - `GlooTest` logically tests every test using `gloo`, `NCCLTest` with NCCL, etc.
  - This allows some niceties, like marking once that we want to skip all NCCL tests 
- Adds `shutdown()` and garbage collection etc. to avoid extraneous messages & warnings like
```
Traceback (most recent call last):
  File "/home/allencwang/workspace/torchft/torchft/process_group.py", line 824, in _future_handler
    cmd = future_queue.get(timeout=timedelta(seconds=10.0))
  File "/home/allencwang/workspace/torchft/torchft/multiprocessing.py", line 45, in get
    raise RuntimeError(f"process is not alive {self._p.exitcode}")
RuntimeError: process is not alive -15
[rank0]:[W207 10:50:12.933128109 CudaIPCTypes.cpp:16] Producer process has been terminated before all shared CUDA tensors released. See Note [Sharing CUDA tensors]
```

# Why is this needed?
As part of https://github.com/pytorch/torchft/pull/102, I noticed that there were some mismatches between which collectives ran on which backends (matrix is [here](https://pytorch.org/docs/stable/distributed.html#backends)). Therefore this logical grouping of tests by backend allows us to define which collectives should be tested explicitly

### DIFF
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -76,43 +76,45 @@ def run_collectives(
     tensor_list = [torch.empty_like(input_tensor)]
 
     works = []
-    input_tensors = []
+    tensors_to_check_list = []
 
     if "allreduce" in collectives:
         works += [
             pg.allreduce([input_tensor], AllreduceOptions()),
             pg.allreduce([input_tensor], ReduceOp.SUM),
         ]
-        input_tensors += [input_tensor, input_tensor]
+        tensors_to_check_list += [input_tensor, input_tensor]
     elif "allgather" in collectives:
         works += [pg.allgather(output_tensors, [input_tensor], AllgatherOptions())]
-        input_tensors += [(output_tensors, input_tensor)]
+        tensors_to_check_list += [(output_tensors, input_tensor)]
     elif "broadcast" in collectives:
         works += [pg.broadcast(tensor_list, BroadcastOptions())]
-        input_tensors += [tensor_list]
+        tensors_to_check_list += [tensor_list]
     elif "broadcast_one" in collectives:
         works += [pg.broadcast_one(input_tensor, 0)]
-        input_tensors += [input_tensor]
+        tensors_to_check_list += [input_tensor]
 
-    def check_tensors(input_tensors: Union[torch.Tensor, List[torch.Tensor]]) -> None:
-        """Recursively check tensors for input_tensors shape and dtype."""
-        if isinstance(input_tensors, torch.Tensor):
+    def check_tensors(
+        tensors_to_check: Union[torch.Tensor, List[torch.Tensor]]
+    ) -> None:
+        """Recursively check tensors for tensors_to_check shape and dtype."""
+        if isinstance(tensors_to_check, torch.Tensor):
             assert (
-                input_tensors.dtype == dtype
-            ), f"Output dtype mismatch: {input_tensors.dtype} != {dtype}"
+                tensors_to_check.dtype == dtype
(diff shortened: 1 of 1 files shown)


torchft/process_group_test.py
```diff
@@ -391,18 +394,19 @@ def test_error_swallowing_process_group_wrapper(self) -> None:
         wrapper = ErrorSwallowingProcessGroupWrapper(pg)
         self.assertIs(wrapper.parent, pg)
 
-        works = _test_pg(wrapper)
+        works = run_collective(pg=wrapper, collective="allreduce")
```
d4l3k, line 397: this seems like a pretty big decrease in coverage?

torchft/process_group_test.py
```diff
@@ -60,24 +61,46 @@ def dummy_init_pg() -> None:
...
+def run_collective(
+    pg: ProcessGroup,
+    collective: str,
+    example_tensor: torch.Tensor = torch.randn((2, 3), dtype=torch.float32),
+) -> Dict[str, dist._Work]:
+    """Run a single collective."""
+    shape: torch.Size = example_tensor.shape
+    dtype: torch.dtype = example_tensor.dtype
+    coll = getattr(pg, collective)
+    args_list = _build_args(pg=pg, collective=collective, example_tensor=example_tensor)
```
d4l3k, line 101: What's the intention behind pulling this out? I'm not really convinced that this makes it all that much cleaner

In some ways I think I'd prefer if we got rid of the arg generation and instead flatten this out with direct calls i.e.

```
if collective == "allreduce":
    work = pg.allreduce(...)

work.wait()
...
```

torchft/process_group_test.py
```diff
@@ -87,98 +110,173 @@ def check_tensors(arg: Any) -> None:  # pyre-ignore[2]
...
-        store = TCPStore(
+    @parameterized.expand(collectives)
+    def test_baby_nccl_apis(self, collective: str) -> None:
+        # set to 1 if more than >=2 gpus
+        device_id = 1 % torch.cuda.device_count()
+        torch.cuda.set_device(device_id)
+
+        pg = ProcessGroupBabyNCCL(timeout=timedelta(seconds=10))
+        try:
+            pg.configure(self.store_addr, 0, 1)
```
d4l3k, line 178: This seems really slow -- how fast does this run? Launching the subprocess is pretty slow so would actually prefer to run these all on the same PG

If you want prettier printing we can use subtests?

i.e.

```
for collective in collectives:
    with self.subTest(collective=collective):
        ...
```

d4l3k (changes requested):

torchft/process_group_test.py
allenwang28: Good callout, with parameterized it took ~36s, without it took ~16s. I've removed parameterized for now.

torchft/process_group_test.py
allenwang28: Added functionality back

allenwang28:
updates in #102 are likely cleaner, so I am going to deprecate this PR

PR #108, opened 2025-02-11
Adds more collectives to ProcessGroups

allenwang28:
# What does this PR do?
Continuation of work for https://github.com/pytorch/torchft/issues/97.

This PR:
- Adds more collectives:
  - allreduce_coalesced
  - alltoall_base
  - barrier
  - reduce_scatter
  - send/recv
- Extends `process_group_test.py` to accommodate the above collectives

Note - the `process_group_test` adds several new tests, increasing the test time from 16s to 22s. Not sure how avoidable this is - a prior version of this PR took 36s!

# Concerns and possible follow up actions

## Missing ops in backends
Notably `allgather_into_tensor_coalesced` and `reduce_scatter_tensor_coalesced` were not added in this PR.  

It was part of the plan, but started seeing an error such as:
```
E       AttributeError: 'torch._C._distributed_c10d.ProcessGroupNCCL' object has no attribute 'allgather_into_tensor_coalesced'
```

This was confusing, but surely enough this is true, if we do `dir(pg)` we see:
```
['NCCLConfig', 'Options', '__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '_add_ephemeral_timeout', '_allgather_base', '_end_coalescing', '_get_backend_name', '_get_sequence_number_for_group', '_group_end', '_group_start', '_is_initialized', '_pybind11_conduit_v1_', '_reduce_scatter_base', '_set_default_timeout', '_set_sequence_number_for_group', '_shutdown', '_start_coalescing', '_verify_work_timeout', 'abort', 'allgather', 'allgather_coalesced', 'allreduce', 'allreduce_coalesced', 'alltoall', 'alltoall_base', 'barrier', 'bound_device_id', 'broadcast', 'comm_split_count', 'deregister_mem_pool', 'eager_connect_single_device', 'gather', 'monitored_barrier', 'name', 'options', 'perform_nocolor_split', 'rank', 'recv', 'recv_anysource', 'reduce', 'reduce_scatter', 'register_mem_pool', 'scatter', 'send', 'size', 'supports_splitting', 'uid']
```

After some digging I realized it was because `ProcessGroupNCCL` (and other pg backends) inherit collectives that are defined [here](https://github.com/pytorch/pytorch/blob/e3839bd60365d97f2509671fd4d0f928af0f6ae1/torch/csrc/distributed/c10d/init.cpp#L2475), which does not include `allgather_into_tensor_coalesced` or `reduce_scatter_tensor_coalesced`. This is confusing since we know that e.g. `allgather_into_tensor_coalesced` [is implemented](https://github.com/pytorch/pytorch/blob/e3839bd60365d97f2509671fd4d0f928af0f6ae1/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp#L667-L670) for the `ProcessGroupNCCL` backend.

There are likely several possible resolutions for this, so will defer for now for a future decision.

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -185,7 +185,7 @@ class ProcessGroup(BaseProcessGroup):
     def reduce_scatter(
         self,
         output_tensors: List[torch.Tensor],
-        input_tensors: List[torch.Tensor],
+        input_tensors: List[List[torch.Tensor]],
         opts: ReduceScatterOptions,
     ) -> Work:
         """
@@ -306,7 +306,7 @@ class ProcessGroupWrapper(ProcessGroup):
     def reduce_scatter(
         self,
         output_tensors: List[torch.Tensor],
-        input_tensors: List[torch.Tensor],
+        input_tensors: List[List[torch.Tensor]],
         opts: object,
     ) -> Work:
         return self.parent.reduce_scatter(output_tensors, input_tensors, opts)
@@ -424,10 +424,10 @@ class ProcessGroupDummy(ProcessGroup):
     def reduce_scatter(
         self,
         output_tensors: List[torch.Tensor],
-        input_tensors: List[torch.Tensor],
+        input_tensors: List[List[torch.Tensor]],
         opts: object,
     ) -> Work:
-        for o, i in zip(output_tensors, input_tensors):
+        for o, i in zip(output_tensors, input_tensors[0]):
             o.copy_(i)
 
         res = _DummyWork(output_tensors)
@@ -1013,7 +1013,6 @@ class ProcessGroupBaby(ProcessGroup):
             for tensor in tensor_list:
                 if not tensor.is_shared():
                     tensor.share_memory_()
-
         return self._run_func("reduce_scatter", output_tensors, input_tensors, opts)
 
     def size(self) -> int:
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -58,6 +58,7 @@ from torch.distributed.distributed_c10d import (
     AllreduceOptions,
     BroadcastOptions,
     ReduceOp,
+    ReduceScatterOptions,
     Work,
 )
 from torch.futures import Future
@@ -180,6 +181,20 @@ class ProcessGroup(BaseProcessGroup):
         opts.rootRank = root
         return self.broadcast([tensor], opts)
 
+    # pyre-fixme[14]: inconsistent override
+    def reduce_scatter(
+        self,
+        output_tensors: List[torch.Tensor],
+        input_tensors: List[torch.Tensor],
+        opts: ReduceScatterOptions,
+    ) -> Work:
+        """
+        Reduces, then scatters a list of tensors to all processes in a group.
+
+        See torch.distributed.reduce_scatter for more details.
+        """
+        raise NotImplementedError("not implemented")
+
     def size(self) -> int:
         raise NotImplementedError("not implemented")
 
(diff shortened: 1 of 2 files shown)


torchft/process_group_test.py
```diff
@@ -122,6 +137,122 @@ def check_tensors(arg: Any) -> None:  # pyre-ignore[2]
     return works
 
 
+def _test_multi_pg(pg: ProcessGroup, rank: int, tensor: torch.Tensor) -> None:
+    """
+    Helper function to test a set of collective operations in settings with multiple
```
rohan-varma, line 142: IMO I'd split these up into individual collective tests, so that it can be easier isolated when there are issues + users will be able to clearly understand which collectives are supported on which fault tolerant process groups.

torchft/process_group_test.py
```diff
@@ -122,6 +137,122 @@ def check_tensors(arg: Any) -> None:  # pyre-ignore[2]
...
+    )
+
+    # Test barrier
+    barrier_work = pg.barrier(BarrierOptions())
+    barrier_work.wait()
+
+    # Test send/recv
+    if rank == 0:
+        send_tensor = tensor.clone()
+        send_work = pg.send([send_tensor], 1, 0)
```
rohan-varma, line 212: Could we also test the actual fault tolerance, i.e. sender fails, sender succeeds but receiver fails, and verify that an appropriate exception / timeout is returned back to the user process? 

And ideally, this is then retried with success after the PG gets reconfigured.

torchft/process_group_test.py
allenwang28: Yeah I agree with the sentiment! I previously tried something like this in https://github.com/pytorch/torchft/pull/103 with `_test_pg`, but this introduced made the execution time much longer since there were now (# of collectives) times as many process groups that need to be spun up and torn down. That approach also added in more complexities that I wasn't comfortable with.

Ideally I could represent it like this:
```
class MultiPgTest:
  def test_all_gather(self):
    tensor_list = ...
    ...
  
  def test_allreduce(self):
    ...
```

but I'm not sure how I could re-use process groups in this way. Open to any suggestions though!

rohan-varma: @allenwang28 Couldn't you just create the process groups & subprocesses in a `setupClass` method which would allow them to be re-used across unittests? 

TestDistBackend and MultiProcessTestCase in PT-D might provide some pointers for doing something like this: https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/distributed/distributed_test.py#L566

allenwang28:
@rohan-varma - these are great suggestions and I agree that these enhancements would improve the testing within torchft.  However this is out of scope for this specific PR so I have created https://github.com/pytorch/torchft/issues/109 to track this request. Thank you for the great feedback!

torchft/process_group_test.py
wconstab: FYI there is a test class called `MultiProcContinuousTest` defined in the same test utils file as MultiProcesTestCase that shares a PG across test instances.  It requires having __main__ defined differently for that test file and isn't compatible with hahving MultiProcesTestCase instnaces inside the same file currently, but it is in use in a number of pt-d tests bc it saves a lot of time

torchft/process_group.py
```diff
@@ -1012,6 +1118,56 @@ def allgather(
...
         return self._run_func("allgather", output_tensors, input_tensor, opts)
 
+    def allreduce(
+        self,
+        tensors: List[torch.Tensor],
+        opts: Union[dist.AllreduceOptions, dist.ReduceOp],
+    ) -> Work:
+        assert isinstance(tensors, list), "input must be list"
+
+        for tensor in tensors:
```
wconstab, line 1128: might be worth adding helper methods for common patterns like 
`maybe_share_tensors(tensor_or_tensors)`

`assert_list(tensors)`
etc?

allenwang28: yea that's clean, added

torchft/process_group_test.py
allenwang28: I tried adding in `subTest` into `_test_multi_pg` to see if it would delineate the scenarios being tested but unfortunately it doesn't seem to play well with `ThreadPoolExecutor`. `subTest` could be a good approach for `_test_pg` though.

PR #110, opened 2025-02-13
checkpointing/PGTransport: add NCCL/Gloo transport for checkpoints

d4l3k:
This adds a ProcessGroup based CheckpointTransport to allow for transferring a state_dict via the backend network.

This supports NCCL with cuda devices and Gloo on CPU devices. DTensor is supported but other tensor subclasses will likely error.

This is cleaned up code from #104 

Additional changes:

* `ProcessGroupBaby`: pass the `timeout` parameter to the subprocess so we can catch NCCL timeouts
* refactored http_transport_test

### Sender

1. preprocess the state_dict using pytree to separate the tensors from python objects
2. serialize the tensor metadata and the non-python objects using pickle
3. send the metadata size via send/recv
4. send the pickled buffer via send/recv
5. send each tensor via send/recv

These sends will send to each of the receiving peers simultaneously.

### Receiver

Receiving is largely the same. Notably when receiving each tensor, a buffer is allocated on the device, received and then transferred to CPU to prevent CUDA ooms. Allocating and transferring is significantly slower than doing and inplace receive and is something we should fix in a follow up PR.

# Test plan:

added a new shared multi rank recovery test and enabled it for PGTransport w/ NCCL+Gloo and the existing HTTPTransport

```
pytest torchft/checkpointing/
```

### DIFF
--- a/torchft/checkpointing/pg_transport.py
+++ b/torchft/checkpointing/pg_transport.py
@@ -1,0 +1,247 @@
+import logging
+import pickle
+import time
+from contextlib import contextmanager
+from dataclasses import dataclass
+from datetime import timedelta
+from typing import Generator, List, Tuple, TypeVar, Union, cast
+
+import torch
+from torch.distributed import Work
+from torch.distributed.tensor import DTensor, _DTensorSpec
+from torch.utils._pytree import TreeSpec, tree_flatten, tree_unflatten
+
+from torchft.checkpointing.transport import CheckpointTransport
+from torchft.process_group import ProcessGroup
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+
+@dataclass
+class _TensorMeta:
+    """
+    This is the metadata for a tensor that is used to transfer checkpoints.
+    It contains the shape, the dtype, the storage offset and the stride of the
+    tensor.
+
+    This must be pickleable so that it can be sent over the wire.
+    """
+
+    shape: torch.Size
+    dtype: torch.dtype
+    storage_offset: int
+    stride: Tuple[int, ...]
+    nbytes: int
+
+
+@dataclass
+class _DTensorMeta:
+    """
+    This is the metadata for a DTensor that is used to transfer checkpoints.
+    It contains the metadata for the local tensor and the spec of the DTensor.
+
+    This must be pickleable so that it can be sent over the wire.
+    """
+
+    local: _TensorMeta
+    spec: _DTensorSpec
+
+
+@dataclass
+class _StateDictMeta:
+    """
+    This is the metadata for a state dict that is used to transfer checkpoints.
+    It contains the step, the pytree spec of the state dict and the metadata for
+    each tensor in the state dict.
+
+    This must be pickleable so that it can be sent over the wire.
+
+    Args:
+        step: the step of the checkpoint to verify consistency
+        treespec: the pytree spec of the state dict
+        non_tensor_leaves: the metadata for each tensor in the state dict and any
+            non-tensor leaves in the state dict
+    """
+
+    step: int
+    treespec: TreeSpec
+    non_tensor_leaves: List[Union[object, _TensorMeta, _DTensorMeta]]
+
+
+@contextmanager
+def _timeit(name: str) -> Generator[None, None, None]:
+    start = time.perf_counter()
+    yield
+    dur = time.perf_counter() - start
+    logger.info(f"{name} took {dur}s")
+
+
+def _prepare_tensor(tensor: torch.Tensor) -> Tuple[torch.Tensor, _TensorMeta]:
+    return (
+        _cast_tensor(tensor, torch.uint8),
+        _TensorMeta(
+            shape=tensor.shape,
+            dtype=tensor.dtype,
+            storage_offset=cast(int, tensor.storage_offset()),
+            stride=tensor.stride(),
+            nbytes=tensor.untyped_storage().nbytes(),
+        ),
+    )
+
+
+def _prepare_state_dict(
+    state_dict: object,
+    step: int,
+    device: torch.device,
+) -> Tuple[_StateDictMeta, List[torch.Tensor]]:
+    leaves, treespec = tree_flatten(state_dict)
+
+    non_tensor_leaves = []
+    tensors = []
+    for v in leaves:
+        if isinstance(v, DTensor):
+            tensor, tensor_meta = _prepare_tensor(v._local_tensor)
+
+            tensors.append(tensor)
+
+            non_tensor_leaves.append(
+                _DTensorMeta(
+                    local=tensor_meta,
+                    spec=v._spec,
+                )
+            )
+        elif isinstance(v, torch.Tensor):
+            tensor, tensor_meta = _prepare_tensor(v)
+            tensors.append(tensor)
+            non_tensor_leaves.append(tensor_meta)
+        else:
+            non_tensor_leaves.append(v)
+
+    return (
+        _StateDictMeta(
+            step=step,
+            treespec=treespec,
+            non_tensor_leaves=non_tensor_leaves,
+        ),
+        tensors,
+    )
+
+
+def _cast_tensor(tensor: torch.Tensor, dtype: torch.dtype) -> torch.Tensor:
+    """
+    Casts the underlying storage to a tensor of the given dtype.
+
+    The returned tensor will be of size ``storage.nbytes``.
+
+    This works for all datatypes and supports strided/offset tensors with the
+    caveat that the cast tensor may be larger than the original tensor due to
+    the differences in striding.
+    """
+    storage = tensor.untyped_storage()
+    ret = torch.tensor(storage, dtype=dtype, device=tensor.device)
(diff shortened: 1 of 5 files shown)


torchft/checkpointing/transport_test.py
```diff
@@ -0,0 +1,147 @@
...
+    self: TestCase,
+    init_transport: Callable[[int, int], CheckpointTransport[Dict[str, object]]],
+    device: torch.device,
+) -> None:
+    """
+    This runs multi node recovery tests for a given transport function.
+
+    This tests up to 3 nodes.
+    """
+    WORLD_SIZE: int = 3
```
allenwang28, line 40: nit: should this be a parameter instead? The function def implies that it's configurable

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+def _timeit(name: str) -> Generator[None, None, None]:
+    start = time.perf_counter()
+    yield
+    dur = time.perf_counter() - start
+    logger.info(f"{name} took {dur}s")
+
+
+def _prepare_tensor(tensor: torch.Tensor) -> Tuple[torch.Tensor, _TensorMeta]:
+    return (
+        _cast_tensor(tensor, torch.uint8),
```
allenwang28, line 55: IIUC, we're casting to uint8 to reduce memory pressure / speed up the transfer, but should we be concerned about any precision loss?

I see that `transport_test.py` verifies closeness/correctness through `run_multi_recovery_test`, but it isn't making sense to me!

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+    ) -> None:
+        self._work: List[Work] = []
+        self._pg = pg
+        self._timeout = timeout
+        self._device = device
+
+    def metadata(self) -> str:
+        return "<n/a>"
+
+    def disallow_checkpoint(self) -> None:
```
allenwang28, line 143: should this be implemented?

allenwang28 (commented):
this is really cool work!

torchft/checkpointing/pg_transport.py
daulet-askarov: This just reinterprets the tensor as a bunch of bytes (hence the uint8) backed by the same UntypedStorage range of bytes. No bytes modified, so no loss of precision:
https://pytorch.org/docs/stable/storage.html#untyped-storage-api

torchft/checkpointing/pg_transport.py
d4l3k: We don't support any async/out of band calls via PG so nothing needs to be done here

For HTTP we need to this to avoid serving a checkpoint during optimizer step but since send is synchronous in PG we don't need any additional synchronization

torchft/checkpointing/transport_test.py
d4l3k: This test explicitly tests sending at sizes 2/3 and will error out if this changes, I can update the comment

torchft/checkpointing/pg_transport.py
d4l3k: In most cases it doesn't matter and should result in a byte identical output.

There's a few trade-offs:

* Pro: not all tensor types are supported by NCCL `torch.uint16` for instance can't be sent via nccl so doing this cast to uint8 allows us to support any dtype
* Con: arguably it's better to use the non-storage option to avoid sending duplicate/extra bytes for strided/offset tensors. If you have two tensors sharing the same underlying storage in this implementation we end up sending twice as much data

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+                work.append(self._pg.send([buf_t], dst_rank, tag=2))
+
+        with _timeit("send tensors"):
+            for i, t in enumerate(tensors):
+                t = t.to(self._device)
+                for dst_rank in dst_ranks:
+                    work.append(self._pg.send([t], dst_rank, tag=3 + i))
+
+                # allow 3 concurrent transfers at a time
+                while len(work) > (3 * len(dst_ranks)):
```
fegin, line 168: Is this used to avoid OOM?

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+        def recv(v: _TensorMeta) -> torch.Tensor:
+            nonlocal i
+
+            t = torch.empty(v.nbytes, dtype=torch.uint8, device=self._device)
+            # TODO: parallelize receives
+            self._pg.recv([t], src_rank, tag=3 + i).wait(timeout)
+            i += 1
+
+            # TODO: allow in place receives to avoid having to copy to cpu to
+            # avoid OOMs
```
fegin, line 200: We should be able to avoid transferring TensorMeta and DTensorMeta and avoid `to(cpu)` if we can first call `state_dict()` to get the `state_dict` structure and traverse the `state_dict` and send/recv the tensor directly.

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@dataclass
+class _DTensorMeta:
+    local: _TensorMeta
+    spec: object
+
+
+@dataclass
+class _StateDictMeta:
```
H-Huang, line 38: I was thinking the `_StateDictMeta` and dataclasses could probably use some more comments since that's pretty important in determining how we serialize / deserialize and being able to update them. I am kinda curious how DCP handles this metadata when it transfers and if we have existing infra we can use? @fegin do you know?

H-Huang (approved):
Overall looks pretty good! I had some comments and questions mostly about around readability and existing infra, but otherwise looks solid enough to land

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+            # TODO: parallelize receives
+            self._pg.recv([t], src_rank, tag=3 + i).wait(timeout)
+            i += 1
+
+            # TODO: allow in place receives to avoid having to copy to cpu to
+            # avoid OOMs
+            t = t.cpu()
+
+            return torch.as_strided(
+                _cast_tensor(t, v.dtype),
```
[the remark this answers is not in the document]
d4l3k: done

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+            tensor, tensor_meta = _prepare_tensor(v)
+            tensors.append(tensor)
+            non_tensors.append(tensor_meta)
+            tensor_metas.append(tensor_meta)
+        else:
+            non_tensors.append(v)
+
+    total_size = sum(t.nbytes for t in tensors)
+
+    dur = time.perf_counter() - start
```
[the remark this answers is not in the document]
d4l3k: done

torchft/checkpointing/pg_transport.py
d4l3k: `.view` doesn't work for strided tensors -- I'm not sure we need to support those but I think I'll leave it as is for now

torchft/checkpointing/pg_transport.py
d4l3k: Yes, it's to avoid OOM when transferring between devices

torchft/checkpointing/pg_transport.py
d4l3k: Yeah we totally can do that -- I wanted to make that refactor in a follow up PR since I still need to figure out how to do that

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+T = TypeVar("T")
+
+
+@dataclass
+class _TensorMeta:
+    shape: torch.Size
+    dtype: torch.dtype
+    storage_offset: int
+    stride: Tuple[int, ...]
+    nbytes: int
```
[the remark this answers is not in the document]
d4l3k: @H-Huang do you know how quantized information is stored? Is it a different tensor subclass or just packed into the storage?

torchft/checkpointing/pg_transport.py
```diff
@@ -0,0 +1,221 @@
...
+        pass
+
+    def send_checkpoint(
+        self, dst_ranks: List[int], step: int, state_dict: T, timeout: timedelta
+    ) -> None:
+        meta, tensors = _prepare_state_dict(state_dict, step, device=self._device)
+
+        work = []
+
+        with _timeit("send pickle"):
```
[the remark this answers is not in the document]
d4l3k: Thanks! Yeah, I missed that before -- this is good to know though for right now I think this implementation is a bit more performant since it sends the same data to multiple receivers

If we wrap the underlying PG we also should be able to use the broadcast_object_list variant which should give us best of both worlds

torchft/checkpointing/pg_transport.py
d4l3k: Added comments and updated field names to make it clearer

fegin: @H-Huang similar to what is does here. Some logic is the same and we can see if we want to unify the code in the future.

PR #111, opened 2025-02-18
Implementing bucketized model averaging for LocalSGD

Krishn1412:
1) Created a flat buffer of the total size of the gradients.
2) Packed all the gradients in the buffer.
3) Performed all_reduce on the buffer and then unpacked it into the original tenors.


torchft/local_sgd.py
```diff
@@ -223,17 +224,35 @@ def _perform_sync(self) -> None:
...
-            work.wait()
+
+        grads = [p.grad for p in self._model.parameters() if p.grad is not None]
+
+        if not grads:
+            return  # No gradients to process
+
+        # Compute total size and allocate a flat buffer for all gradients
+        total_size = sum(g.numel() for g in grads)
+        flat_buffer = torch.zeros(total_size, dtype=grads[0].dtype, device=grads[0].device)
```
d4l3k, line 240: What I had in mind for "bucketizing" was similar to how DDP workers -- i.e. bucketize into tensors of say ~32MB and run the operations on those. To then save memory PyTorch DDP backs the original tensors with a view into the bucketized tensor storage. I'm not quite sure how manipulating the original storage would work interact with FSDP

Since this allocates extra memory I think we may want to have an option to preserve the original behavior. If we want to keep this implementation as an intermediate option we should probably pull these out into helper methods and add a flag

d4l3k (changes requested):
To land this algorithm we should add a flag to control new vs old behavior

Also need some unit tests

Thanks for taking a stab at this!

torchft/local_sgd.py
Krishn1412: Sure thing @d4l3k, I'll add a flag and unit tests for this. Should I also change the size of the buffer to a constant like 32 MB?

d4l3k: @Krishn1412 yeah, that would be best

Stretch goal would be to add a small benchmark and tune the performance as well but can do in a follow up PR as well

torchft/local_sgd_test.py
```diff
@@ -144,3 +144,57 @@ def test_diloco_healthy(self) -> None:
 
             outer_opt_state = outer_optimizer.state_dict()
             self.assertEqual(len(outer_opt_state["state"]), parameter_count)
+
+    def test_diloco_without_bucketization(self):
```
d4l3k, line 148: Can we make this parameterized instead of duplicating the test twice?

torchft/local_sgd.py
```diff
@@ -226,14 +230,64 @@ def _perform_sync(self) -> None:
...
+            if p.grad is None:
+                continue
             work = self._manager.allreduce(p.grad)
             works.append(work)
-        # Wait for all allreduce operations to complete
+
         for work in works:
             work.wait()
+
+    def _allreduce_bucketized(self) -> None:
```
d4l3k, line 254: can we add a correctness test for this bucketization logic?

torchft/local_sgd.py
```diff
@@ -182,6 +183,8 @@ class DiLoCo(LocalSGD):
     diloco: https://arxiv.org/pdf/2311.08105
     """
 
+    BUCKET_SIZE_BYTES = 32 * 1024 * 1024
```
d4l3k, line 186: Can we make this configurable parameter named `bucket_cap_mb` to match DDP? 

https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html

torchft/local_sgd.py
```diff
@@ -191,6 +194,7 @@ def __init__(
         sync_every: int,
         backup_device: Optional[torch.device] = None,
         pin_memory: bool = True,
+        use_bucketization=False,
```
d4l3k, line 197: nit: missing types

d4l3k (changes requested):
Code is looking pretty good -- just need a few cleanups on lint + some unit tests

d4l3k (approved):
LGTM

d4l3k:
@Krishn1412 there's some big merge conflicts now -- would you mind rebasing on the latest changes?

Would also be a bit simplified if we pull the bucketizing algorithm into a dedicated function that can be reused in different spots

Krishn1412:
@d4l3k I have rebased and resolved the conflicts. I will create the bucketization logic in a specific function. Is there a utils file of some sort where you want this?

Krishn1412:
Hey @d4l3k , I have been trying to run the python unit tests locally but I am getting this error,

`ImportError while importing test module '/Users/krishnparasar/Desktop/projects/torchft/torchft/process_group_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/importlib/__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
torchft/__init__.py:8: in <module>
    from torchft.ddp import DistributedDataParallel
torchft/ddp.py:26: in <module>
    from torchft.process_group import ProcessGroup, ProcessGroupDummy, ProcessGroupGloo
torchft/process_group.py:45: in <module>
    from torch.distributed import (
E   ImportError: cannot import name 'ProcessGroupNCCL' from 'torch.distributed'`

Is this a known issue? I was thinking to move the NCCL related code to a different file or add some sort of flag.

d4l3k:
@Krishn1412 this is a known issue that we don't support Mac. See #136 for more details. 

Happy to have mac support added, I just don't have access to a Mac computer right now

Krishn1412:
Cool @d4l3k , I'll try out the hack mentioned there.

Krishn1412:
hey @d4l3k , thanks for adding the Mac support. It's easier for me to contribute now since I have moved to Mac. Can you take a look at the PR now.

Krishn1412:
hey @d4l3k, did you take a look at the PR?

d4l3k:
@Krishn1412 sorry for the delay in review -- code looks good but seems like there's a merge conflict now

Krishn1412:
@d4l3k, I have resolved the conflict.

d4l3k:
@Krishn1412 thanks for all your work and patience getting this in!

Krishn1412:
@d4l3k appreciate it, looking forward to more collaboration and contribution.

PR #114, opened 2025-02-21
Adds support for `allgather_into_tensor_coalesced` and `reduce_scatter_tensor_coalesced` 

allenwang28:
Closes out https://github.com/pytorch/torchft/issues/97

# What does this PR do?
This PR introduces support for `allgather_into_tensor_coalesced` and `reduce_scatter_tensor_coalesced`, similar to other supported collectives. This is a follow up to https://github.com/pytorch/torchft/pull/108

As noted in https://github.com/pytorch/torchft/pull/108, `allgather_into_tensor_coalesced` and `reduce_scatter_tensor_coalesced` are supported in the base `ProcessGroup` but not within `ProcessGroupNCCL`. 

@d4l3k  pointed out we may be able to create a regular `CppProcessGroup` and register a backend, rather than return the backend which is what we were doing before, which seemed to work!

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -399,6 +399,7 @@ class ProcessGroupWrapper(ProcessGroup):
         )
 
     def barrier(self, opts: BarrierOptions) -> Work:
+        print(type(self.parent))
         return self.parent.barrier(opts)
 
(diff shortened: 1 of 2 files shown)


d4l3k (approved):
LGTM, thanks!

PR #115, opened 2025-02-22
Rename torchft to _torchft in project.scripts section

mreso:
Seems like #84 renamed the module but it was not updated in the `pyproject.toml`.

Got `ModuleNotFoundError: No module named 'torchft.torchft'`


d4l3k (approved):
LGTM

d4l3k:
We should probably add a test for this

PR #120, opened 2025-03-01
Clean up local_sgd and diloco

H-Huang:
Changes:
- Remove backup parameters from local_sgd
- clean up tests
- Make DiLoCo not a subclass of local_sgd (there is some code repetition, but I think this is okay if we want to use these as an example for using our torchft APIs)

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -10,7 +10,7 @@ This module implements a fault tolerant version of LocalSGD and related methods.
 """
 import logging
 from types import TracebackType
-from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, Type
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type
 
 import torch
 from torch import nn, optim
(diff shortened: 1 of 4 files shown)


torchft/local_sgd.py
```diff
@@ -151,30 +121,30 @@ def sync(self) -> None:
...
+        """
         works = []
-
+        averaged_parameters = []
         for p in self._model.parameters():
-            # TODO: bucketize parameters
-            works.append(self._manager.allreduce(p.data.detach()))
-
+            # Create a new tensor to store the averaged parameter
+            avg_param = p.data.clone()
```
d4l3k, line 139: Should we be checking/setting p.data.grad to None to minimize memory impact from the copy here?

d4l3k (approved):
LGTM

torchft/local_sgd.py
H-Huang: I think `p.data.grad` should always be None, but I can add a line to ensure it is `None`. I think only `p.grad` gets set, not sure when `p.data.grad` would get updated

PR #122, opened 2025-03-05
Improve log readability

H-Huang:
Couple of changes to logging:
- Use a different logging crate
- Allow log level to be set with `RUST_LOG` env
- Add `log_if_changed` to lighthouse.rs so it will only log if there is a message change. The quorum tick is pretty frequent by default
- Preprend replica_id to manager.rs logs by adding a new logging macro `info_with_replica`

Open to suggestion / changes.

Previous log example:
image

full paste: P1747150901

New log example:
image

full paste: P1747151799

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -37,6 +37,16 @@ use log::{info, warn};
 #[cfg(test)]
 use std::{println as info, println as warn};
 
+macro_rules! info_with_replica {
+    ($replica_id:expr, $($arg:tt)*) => {
+        info!(
+            "[Replica {}] {}",
+            $replica_id,
+            format!($($arg)*)
+        );
+    };
+}
+
 struct ManagerState {
     checkpoint_metadata: HashMap<i64, String>,
     channel: broadcast::Sender<Quorum>,
@@ -63,7 +73,10 @@ pub async fn manager_client_new(
     addr: String,
     connect_timeout: Duration,
 ) -> Result<ManagerServiceClient<Channel>> {
-    info!("ManagerClient: establishing connection to {}", &addr);
+    info!(
+        "Creating ManagerClient: establishing connection to {}",
+        &addr
+    );
     let conn = connect(addr, connect_timeout).await?;
     Ok(ManagerServiceClient::new(conn))
 }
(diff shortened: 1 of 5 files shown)


torchft/manager.py
```diff
@@ -191,7 +190,7 @@ def __init__(
 
             if replica_id is None:
                 replica_id = ""
-            replica_id = replica_id + str(uuid.uuid4())
```
H-Huang, line 194: Wasn't sure why uuid was added to the replica_id, is this needed?

fegin: ye, I was wondering the same thing. TorchTitan has to keep the original `replicate_id` because this change every time due to `uuid4`.

fegin:
`Allow log level to be set with RUST_LOG env`
^
This is super useful !

torchft/manager.py
d4l3k: We do this as a simplification for the quorum algorithm -- if a worker restarts quickly with the same name we may not detect that the quorum has changed and thus no-reconfiguration of the PGs will occur. Adding a random UUID to the worker means that we will always detect processes restart and thus correctly trigger a reconfiguration

@H-Huang can you add this back and if it's convenient also add a comment explaining why this is the case? Also it would be nice to add a `:` to it so `replica_01234-1234` shows up as `replica_0:1234-1234` . I've been meaning to fix this but haven't had a chance

src/manager.rs
```diff
@@ -37,6 +37,16 @@ use log::{info, warn};
 #[cfg(test)]
 use std::{println as info, println as warn};
 
+macro_rules! info_with_replica {
+    ($replica_id:expr, $($arg:tt)*) => {
+        info!(
+            "[Replica {}] {}",
+            $replica_id,
```
d4l3k, line 44: Can we make this macro access `self.replica_id` instead of taking it as a parameter?

H-Huang: We would have to move `compute_quorum_results()` (https://github.com/pytorch/torchft/blob/082753cbe9705d00f63752e447aacfd6c21ca502/src/manager.rs#L357) into Manager class if we want to pass in the self instance, should we do that?

d4l3k: ahh this is fine then -- was just curious if we could clean it up a bit but this is fine

d4l3k (approved):
LGTM

PR #125, opened 2025-03-07
Set current device for async_quorum and PG process

fegin:


d4l3k (approved):
LGTM, list needs fixing

PR #128, opened 2025-03-10
Add GPU tests for local sgd / diloco

H-Huang:
Update local sgd tests to train the model on GPU and uses `ProcessGroupBabyNCCL` for synchronization


d4l3k (approved):
LGTM!

PR #138, opened 2025-03-17
run CI with torch nightly

d4l3k:
This adds a new CI build that uses torch nightly to make sure we're compatible with trunk.

Nightly doesn't support old versions of glibc/CUDA driver so we needed to make a few upgrades to the runners.

This makes a few changes to CI for compatibility:

* update to linux_job_v2.yaml which uses a newer glibc
* use Python 3.12 since 3.10 has issues with the older glibc
* use CUDA 12.4 for the newer driver
* switch to g5.12xlarge (4 GPU A10G) which are compatible with newer CUDA version and can run all the multi GPU tests

Test plan:

CI

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -9,16 +9,23 @@ on:
 jobs:
   unittest:
     strategy:
+      fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


H-Huang (approved):
LGTM

PR #149, opened 2025-03-25
disable cuda tests in local_sgd_integ_test

H-Huang:
`ProcessGroupBabyNCCL` has an issue with allreduce race condition that I haven't figured out.  I tried running in processes instead of threads and it works, but I haven't figured out a good way to update the integration tests yet.

The values of the tensor after the allreduce is not the same or expected, e.g.
```
BEFORE average, replica 1:08cb3c4c-365e-4c8f-86d2-87000e78f2ad [model.0.bias tensor([-0.0004,  0.0006, -0.0008], device='cuda:4')]
BEFORE average, replica 0:812b70ca-ce59-43f1-b602-18515f23d808 [model.0.bias tensor([ 0.0003,  0.0002, -0.0008], device='cuda:0')]
AFTER average, replica 1:08cb3c4c-365e-4c8f-86d2-87000e78f2ad [model.0.bias tensor([ 6.1110e-05,  5.4164e-04, -1.2743e-03], device='cuda:4')]
AFTER average, replica 0:812b70ca-ce59-43f1-b602-18515f23d808 [model.0.bias tensor([ 3.0555e-05,  2.7082e-04, -6.3717e-04], device='cuda:0')]
```

Disabling the cuda tests for now so the CI signal is less flaky.

### DIFF
--- a/torchft/local_sgd_integ_test.py
+++ b/torchft/local_sgd_integ_test.py
@@ -197,9 +197,11 @@ def diloco_train_loop(
 
 
 class LocalSGDIntegTest(TestCase):
+    # TODO: race condition due to using NCCL in threads causes manager allreduce to sometimes not be correct
+    # Because of that the test is disabled for cuda
     @parameterized.expand(
         [
-            (True,),
+            # (True,),
             (False,),
         ]
     )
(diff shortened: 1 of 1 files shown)


d4l3k (approved):
Are these tests more reliable using ProcessGroupNCCL?

Will accept to reduce CI failures but we should figure out what's going on here

H-Huang:
`ProcessGroupNCCL` works for me locally, but for some reason hangs in the CI. So there are also issues there.

Issue #140, opened 2025-03-18
support generic quorum api on LighthouseClient


d4l3k:
We want to provide a generic `quorum` primitive on LighthouseClient. Much of the refactoring to expose these clients has been completed but we don't provide a generic version of quorum yet.

This should take in a Python dictionary of values and then return the computed quorum with all values.

Example usage:

```py
client = LighthouseClient("http://...")

quorum = client.quorum(..., data={"my_data": 1234})

for member in quorum:
    print(member.replica_id, member.data)
```

Code pointers: https://github.com/pytorch/torchft/blob/main/src/lib.rs#L142-L150

This has been requested by Slava.

PR #150, opened 2025-03-26
Support generic quorum api on LighthouseClient

fduwjj:
This is to address the issue here: https://github.com/pytorch/torchft/issues/140

generated doc:

image

run the following test:
 pytest torchft/coordination_test.py
pytest torchft/lighthouse_test.py -k test_lighthouse_client_behavior

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ gethostname = "0.5.0"
 log = "0.4.22"
 prost = "0.13.3"
 prost-types = "0.13.3"
-pyo3 = {version = "0.22.3", features = ["extension-module"]}
+pyo3 = {version = "0.24", features = ["extension-module"]}
 rand = "0.8.5"
 slog = "2.7.0"
 slog-stdlog = "4.1.1"
--- a/proto/torchft.proto
+++ b/proto/torchft.proto
@@ -42,6 +42,7 @@ message QuorumMember {
     int64 step = 4;
     uint64 world_size = 5;
     bool shrink_only = 6;
+    // User passing in data stored as JSON string.
     string data = 7;
 }
 
--- a/proto/torchft.proto
+++ b/proto/torchft.proto
@@ -42,6 +42,7 @@ message QuorumMember {
     int64 step = 4;
     uint64 world_size = 5;
     bool shrink_only = 6;
+    string data = 7;
 }
 
 message Quorum {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -417,6 +417,8 @@ struct LighthouseClient {
     runtime: Runtime,
 }
 
+use pyo3::types::PyString;
+use pyo3::types::PyDict;
 #[pymethods]
 impl LighthouseClient {
     #[new]
@@ -437,7 +439,7 @@ impl LighthouseClient {
         })
     }
 
-    fn quorum(
+    fn quorum<'py>(
         &self,
         py: Python<'_>,
         replica_id: String,
@@ -447,10 +449,13 @@ impl LighthouseClient {
         world_size: u64,
         shrink_only: bool,
         timeout: Duration,
-        data: Option<PyObject>,
+        data: Option<&Bound<'_, PyDict>>,
     ) -> Result<PyQuorum, StatusError> {
         let data_string = match data {
-            Some(obj) => obj.extract(py)?,
+            Some(d) => {
+                let py_str: &Bound<PyString> = &d.str()?;
+                py_str.to_str()?.to_owned()
+            }
             None => String::new(),
         };
         py.allow_threads(move || {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,8 +35,7 @@ pub mod torchftpb {
 use crate::torchftpb::lighthouse_service_client::LighthouseServiceClient;
 use crate::torchftpb::manager_service_client::ManagerServiceClient;
 use crate::torchftpb::{
-    CheckpointMetadataRequest, LighthouseQuorumRequest, ManagerQuorumRequest, Quorum, QuorumMember,
-    ShouldCommitRequest,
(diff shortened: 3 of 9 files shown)


Cargo.toml
```diff
@@ -14,7 +14,7 @@ gethostname = "0.5.0"
 log = "0.4.22"
 prost = "0.13.3"
 prost-types = "0.13.3"
-pyo3 = {version = "0.22.3", features = ["extension-module"]}
+pyo3 = {version = "0.24", features = ["extension-module"]}
```
fduwjj, line 17: This is indeed otherwise trying examples in https://pyo3.rs/v0.24.0/. Somehow 0.24 is already the stable version.. Their API is not BC compatible, I spent whole debugging why some APIs are not found in the scope..

src/pytorch-sphinx-theme
d4l3k, line 1: Why do we need this under src? We have this under docs/src/pytorch-sphinx-theme right?

torchft/_torchft.pyi
```diff
@@ -59,3 +60,42 @@ class LighthouseServer:
     ) -> None: ...
     def address(self) -> str: ...
     def shutdown(self) -> None: ...
+
+@dataclass
+class PyQuorumMember:
```
d4l3k, line 65: Can we avoid the Py* prefixes for these? If you're using them in Python you already know they're Python

torchft/lighthouse_test.py
```diff
@@ -1,10 +1,11 @@
 import time
+from datetime import timedelta
 from unittest import TestCase
 
 import torch.distributed as dist
 
 from torchft import Manager, ProcessGroupGloo
-from torchft._torchft import LighthouseServer
+from torchft._torchft import LighthouseClient, LighthouseServer
```
d4l3k, line 8: We should expose the `LighthouseClient` in `torchft.coordination`. That way it'll get documented on the site

proto/torchft.proto
```diff
@@ -42,6 +42,7 @@ message QuorumMember {
     int64 step = 4;
     uint64 world_size = 5;
     bool shrink_only = 6;
+    string data = 7;
```
d4l3k, line 45: can we document this as being JSON data?

src/pytorch-sphinx-theme
fduwjj: It may be generated accidentally.. let me revert this change.

torchft/_torchft.pyi
fduwjj: This is to avoid duplicate naming with the one from protobuffer. Let me try to see if it works or not.

torchft/lighthouse_test.py
fduwjj: sure.

proto/torchft.proto
fduwjj: sure.

torchft/lighthouse_test.py
```diff
@@ -65,3 +66,47 @@ def test_heartbeat_timeout_ms_sanity(self) -> None:
...
+                store_address=f"localhost:{store.port}",
+                step=1,
+                world_size=1,
+                shrink_only=False,
+                timeout=timedelta(seconds=1),
+                data={"my_data": 1234},
+            )
+            assert result is not None
+            assert len(result.participants) == 1
+            for member in result.participants:
```
d4l3k, line 104: can we add some asserts on the types of these?

src/lib.rs
```diff
@@ -482,7 +673,11 @@ fn _torchft(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<ManagerServer>()?;
     m.add_class::<ManagerClient>()?;
     m.add_class::<LighthouseServer>()?;
+    m.add_class::<LighthouseClient>()?;
     m.add_class::<QuorumResult>()?;
+    m.add_class::<PyQuorumMember>()?;
+    m.add_class::<PyQuorum>()?;
+    m.add_class::<PyTimestamp>()?;
```
d4l3k, line 680: Don't we need to rename these fields as well? Are we overriding the Py prefix somewhere?

Would also be good to add a test with asserts on the return types of the objects in Python

d4l3k (approved):
LGTM, thanks for adding this -- just a few small comments

src/lib.rs
```diff
@@ -337,6 +342,186 @@ async fn lighthouse_main_async(opt: lighthouse::LighthouseOpt) -> Result<()> {
...
+
+    fn quorum<'py>(
+        &self,
+        py: Python<'_>,
+        replica_id: String,
+        address: String,
+        store_address: String,
+        step: i64,
+        world_size: u64,
+        shrink_only: bool,
```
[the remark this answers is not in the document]
fduwjj: I see.. Sure, I can make it optional and wrap it with default value.

torchft/lighthouse_test.py
fduwjj: sure.

fduwjj: sure.

src/lib.rs
fduwjj: OK, figured out a way to disambiguate same naming.

src/lib.rs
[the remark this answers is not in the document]
fduwjj: For this one, I think I need to think more on which one we want to make it optional. I will send out a follow-up PR to address this comment.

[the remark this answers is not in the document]
d4l3k: Sounds good, we can do it in a follow up

fduwjj:
failed test is not related.

src/lib.rs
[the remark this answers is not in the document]
fduwjj: follow-up PR here: https://github.com/pytorch/torchft/pull/154

PR #154, opened 2025-04-02
Make params of quorum of lighthouse client optional

fduwjj:
Address the comment in https://github.com/pytorch/torchft/pull/150/files#r2015127529 to make the input params of quorum optional.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,6 @@ use crate::torchftpb::{
 };
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyString};
-use pyo3::wrap_pyfunction;
 
 // Get the number of threads to use for the tokio runtime
 fn num_threads() -> usize {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -501,44 +501,38 @@ impl LighthouseClient {
     ///
     /// Args:
     ///     replica_id (str): The string id of the replica calling quorum.
-    ///     address (Optional[str]): The address of the replica calling quorum.
-    ///     store_address (Optional[str]): The address of the store.
-    ///     step (Optional[int]): The step of the replica calling quorum.
-    ///     world_size (Optional[int]): The world size of the replica calling quorum.
-    ///     shrink_only (Optional[bool]): Whether the quorum is for shrinking only.
-    ///     timeout (Optional[timedelta]): The timeout for quorum.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -501,28 +501,44 @@ impl LighthouseClient {
...
+        address: Option<String>,
+        store_address: Option<String>,
+        step: Option<i64>,
+        world_size: Option<u64>,
+        shrink_only: Option<bool>,
+        timeout: Option<Duration>,
         data: Option<&Bound<'_, PyDict>>,
     ) -> Result<Quorum, StatusError> {
+        let address = address.unwrap_or("localhost".to_string());
+        let store_address = store_address.unwrap_or("localhost".to_string());
```
d4l3k, line 537: It might be better to make these either `""` or `"<not set>"` so they're not accidentally used if they're invalid values

src/lib.rs
```diff
@@ -501,28 +501,44 @@ impl LighthouseClient {
...
     ///     Quorum: Current quorum if successful.
+    #[pyo3(signature = (
+        replica_id,
+        address = None,
+        store_address = None,
+        step = None,
+        world_size = None,
+        shrink_only = None,
+        timeout = None,
+        data = None
```
d4l3k, line 522: should we be setting the defaults here instead of below with the `unwrap_or`?

src/lib.rs
```diff
@@ -501,28 +501,44 @@ impl LighthouseClient {
...
-        step: i64,
-        world_size: u64,
-        shrink_only: bool,
-        timeout: Duration,
+        address: Option<String>,
+        store_address: Option<String>,
+        step: Option<i64>,
+        world_size: Option<u64>,
+        shrink_only: Option<bool>,
+        timeout: Option<Duration>,
```
d4l3k, line 533: timeout is fine to keep as non-optional as it's a pretty important value

d4l3k (approved):
Change seems reasonable but I think we can do some small changes to make it even better

src/lib.rs
fduwjj: sure, let me do that.

src/lib.rs
fduwjj: I see. Then I need to move the param order a bit to make it after `replicate_id`. Let me change that.

PR #156, opened 2025-04-07
Make ProcessGroupNCCL available for import

H-Huang:
including other small fixes to get ft working with titan again


d4l3k (approved):
LGTM, we may need to make this conditional

PR #160, opened 2025-04-10
ci: use torch 2.7 RC builds

d4l3k:
This updates the minimum version of torch to be 2.7.0. There's some important fixes in PyTorch 2.7 that avoid crashes and other slow behavior.

CI has been flaky recently in large part due to segfaults caused by bugs in 2.6.

Test plan:

CI

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -23,6 +23,9 @@ jobs:
 
           sudo apt-get install -y protobuf-compiler
 
+          # use RC build
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/test/cu128
+
           pip install lintrunner lintrunner-adapters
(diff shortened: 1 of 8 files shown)


.github/workflows/lint.yaml
```diff
@@ -23,6 +23,9 @@ jobs:
 
           sudo apt-get install -y protobuf-compiler
 
+          # use RC build
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/test/cu128
```
H-Huang, line 27: does the lint workflow need cuda?

H-Huang (approved):

.github/workflows/lint.yaml
d4l3k: Yes, for ProcessGroupNCCL

Issue #151, opened 2025-03-26
Add max retries to quorum to avoid live locks


d4l3k:
When we hit issues during training we can sometimes hit issues that are not recoverable and consistently fail but we retry. We should add an option to Manager `max_retries` which allows the user to say only retry `should_commit` up to N times before crashing and allowing the underlying scheduler to retry.

This should help avoid live lock behavior.

Relevant code:

* Manager.__init__ https://github.com/pytorch/torchft/blob/main/torchft/manager.py#L88
* Where we should check for should_commit success https://github.com/pytorch/torchft/blob/main/torchft/manager.py#L631
* manager mocked tests https://github.com/pytorch/torchft/blob/main/torchft/manager_test.py

Testing:

This should be fairly straightforward to add a manager test that uses the mocked manager and verifies that after N failed `should_commit` calls it throws an exception.

PR #166, opened 2025-04-16
Add max retries to quorum to avoid live locks

msaroufim:
Fixes https://github.com/pytorch/torchft/issues/151

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -107,6 +107,7 @@ class Manager:
         heartbeat_interval: timedelta = timedelta(milliseconds=100),
         checkpoint_transport: Optional[CheckpointTransport[Dict[str, T]]] = None,
         init_sync: bool = True,
+        max_retries: Optional[int] = None,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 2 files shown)


torchft/manager.py
```diff
@@ -638,6 +648,20 @@ def should_commit(self, timeout: Optional[timedelta] = None) -> bool:
...
+            self._commit_failures = 0  # Reset failure counter on success
+        else:
+            self._commit_failures += 1
+            # Check if we've hit max retries
+            if (
+                self._max_retries is not None
+                and self._commit_failures > self._max_retries
+            ):
+                self._logger.exception(
+                    f"should_commit failed {self._commit_failures} times (max_retries={self._max_retries})"
```
d4l3k, line 660: nit:

```
msg = f"..."
self._logger.exception(msg)
raise RuntimeError(msg)
```

torchft/manager.py
```diff
@@ -670,7 +695,11 @@ def state_dict(self) -> Dict[str, int]:
         Returns:
             the state dict for this manager
         """
-        return {"step": self._step, "batches_committed": self._batches_committed}
+        return {
+            "step": self._step,
+            "batches_committed": self._batches_committed,
+            "commit_failures": self._commit_failures,
```
d4l3k, line 701: This is an interesting case that I hadn't considered -- I think it might be useful since it can avoid out of sync issues but I'm also a tad concerned if we write it out to persistent storage and then load from checkpoint. Fresh startup should likely restart this count

Thoughts on leaving this out for now?

d4l3k (approved):
LGTM, thanks for sending this! Code looks good though one question on checkpointing behavior

Issue #152, opened 2025-03-26
Add watchdog to _TimeoutManager+ProcessGroupNCCL to guarantee fast aborts


d4l3k:
Add a watchdog to the _TimeoutManager so if we ever end up in a spot where the timeout handling hangs (i.e. ncclCommAbort hangs) we could crash the whole program. The timeout handlers in _TimeoutManager should complete promptly.

If we add a watchdog thread to the _TimeoutManager it should automatically give us watchdog behavior on top of ProcessGroupNCCL without adding any additional complexity to that code.

Code pointer: https://github.com/pytorch/torchft/blob/main/torchft/futures.py

PR #167, opened 2025-04-17
[WIP] Watchdog

msaroufim:
Fixes https://github.com/pytorch/torchft/issues/152

Usage

```python
from datetime import timedelta
from torchft.process_group import ProcessGroupNCCL

# Create a process group with watchdog (5 second timeout)
pg = ProcessGroupNCCL(
    timeout=timedelta(seconds=60.0),
    watchdog_timeout=timedelta(seconds=5.0)
)

# To disable the watchdog
pg = ProcessGroupNCCL(watchdog_timeout=None)
```

### DIFF
--- a/docs/source/process_group.rst
+++ b/docs/source/process_group.rst
@@ -10,12 +10,6 @@ The ``ProcessGroupNCCL`` class includes a watchdog feature for fast aborts of NC
 This is particularly useful in distributed training scenarios where NCCL operations might hang 
 indefinitely, causing the entire training job to stall.
 
-Key features:
-  
-* Monitors operations: A background thread tracks all active NCCL operations
-* Fast abort: Promptly aborts operations that exceed the configured timeout
-* Configurable: Can be enabled or disabled with a custom timeout
-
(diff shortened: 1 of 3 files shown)


d4l3k:
@msaroufim thanks for taking a stab at this! I am a bit concerned that this approach adds significant amount of code/complexity to the ProcessGroupNCCL implementation and also makes the watchdog not very reusable.

What I had in mind for this was to add a watchdog to the _TimeoutManager so if we ever end up in a spot where the timeout handling hangs (i.e. ncclCommAbort hangs) we could crash the whole program. The timeout handlers in `_TimeoutManager` should complete promptly.

If we add a watchdog thread to the _TimeoutManager it should automatically give us watchdog behavior on top of ProcessGroupNCCL without adding any additional complexity to that code.

Code pointer: https://github.com/pytorch/torchft/blob/main/torchft/futures.py

Happy to chat more about this offline

msaroufim:
That does sound much simpler lol, lemme close this and redo

PR #168, opened 2025-04-18
Support DTensor params in local_sgd/diloco

H-Huang:
Tested in torchtitan (https://github.com/pytorch/torchtitan/pull/1122) when the model is sharded with FSDP we need to convert the params to local tensors, update, then back to DTensors.

Ideally I would like to add a test in the integration tests but that would require us to set up the device_mesh / FSDP for a model

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -14,6 +14,7 @@ from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type
 
 import torch
 from torch import nn, optim
+from torch.distributed.tensor import DTensor
 from torch.nn.parameter import Parameter
 from torch.optim.optimizer import Optimizer
 from torch.utils.hooks import RemovableHandle
@@ -21,6 +22,20 @@ from torch.utils.hooks import RemovableHandle
 from torchft.manager import Manager
 
 logger: logging.Logger = logging.getLogger(__name__)
+
+
+def extract_local_tensor(t: torch.Tensor) -> torch.Tensor:
+    """
+    Returns a cloned version of the input tensor. If the input tensor is a DTensor,
+    it extracts and clones its local representation.
+    """
+    new_tensor = None
+    if isinstance(t, DTensor):
+        new_tensor = t.to_local().clone()
+    else:
+        new_tensor = t.clone()
+    new_tensor.grad = None
+    return new_tensor
 
 
 class LocalSGD:
@@ -126,7 +141,15 @@ class LocalSGD:
         if self._manager.should_commit():
             # Update the model parameters with the averaged values
             for param, avg_param in zip(self._model.parameters(), averaged_parameters):
-                param.data.copy_(avg_param)
+                if isinstance(param, DTensor):
+                    # we averaged the local version of the tensor so need to copy it back as a DTensor
+                    param.data.copy_(
+                        DTensor.from_local(
+                            avg_param, param.device_mesh, param.placements
+                        )
+                    )
+                else:
+                    param.data.copy_(avg_param)
 
     def _average(self) -> list[torch.Tensor]:
         """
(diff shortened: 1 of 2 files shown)


torchft/local_sgd.py
```diff
@@ -252,8 +286,10 @@ def _perform_sync(self) -> None:
         """
         # Set the .grad field of each parameter to its pseudogradient
         for name, p in self._model.named_parameters():
-            pseudogradient = p.data - self.original_parameters[name]
-            p.grad = pseudogradient
+            local_param = extract_local_tensor(p.data)
+            pseudogradient = local_param - self.original_parameters[name]
+            # TODO: is there something other than _local_tensor?
+            p._local_tensor.grad = pseudogradient
```
H-Huang, line 292: cc @XilunWu, is there a better way to update the gradient of a `DTensor`?

If I try to set p.grad directly I get an error like:
```
    File "/data/users/howardhuang/torchft/torchft/local_sgd.py", line 293, in _perform_sync
      p.grad = pseudogradient
  RuntimeError: attempting to assign a gradient of size '[564, 256]' to a tensor of size '[2256, 256]'. Please ensure that the gradient and the tensor are the same size
```

torchft/local_sgd.py
```diff
@@ -252,8 +286,13 @@ def _perform_sync(self) -> None:
         """
         # Set the .grad field of each parameter to its pseudogradient
         for name, p in self._model.named_parameters():
-            pseudogradient = p.data - self.original_parameters[name]
-            p.grad = pseudogradient
+            local_param = extract_local_tensor(p.data)
+            pseudogradient = local_param - self.original_parameters[name]
+            # TODO: is there something else available other than _local_tensor?
+            if isinstance(p, DTensor):
+                p._local_tensor.grad = pseudogradient
```
H-Huang, line 293: cc @XilunWu, is there a better way to update the gradient of a `DTensor`?

If I try to set p.grad directly I get an error like:
```
    File "/data/users/howardhuang/torchft/torchft/local_sgd.py", line 293, in _perform_sync
      p.grad = pseudogradient
  RuntimeError: attempting to assign a gradient of size '[564, 256]' to a tensor of size '[2256, 256]'. Please ensure that the gradient and the tensor are the same size
```

XilunWu: I think this is the right way. cc @weifengpy

weifengpy: looks right to me

weifengpy: after some discussion, actually `param.grad._local_tensor` might be better, because optimizer always call param.grad regardless of DTensor

H-Huang:
Without the Dtensor conditionals, the operations like `.copy_()` and pseudogradient calculation fail due to incompatible types. The `allreduce` also fails because `got exception in all reduce -- skipping remaining: found no DeviceMesh from dtensor args for c10d.allreduce_.default!`

torchft/local_sgd.py
```diff
@@ -198,13 +222,23 @@ def _save_parameters(self) -> None:
...
-                self.original_parameters[name].copy_(p.data, non_blocking=True)
+                param_to_local = extract_local_tensor(p.data)
+                self.original_parameters[name].copy_(param_to_local, non_blocking=True)
 
     def _restore_parameters(self) -> None:
         with torch.no_grad():
             # TODO: consider running copy on a separate stream
             for name, p in self._model.named_parameters():
-                p.data.copy_(self.original_parameters[name], non_blocking=False)
+                if isinstance(p, DTensor):
```
fegin, line 232: If `p` is a DTensor, does `p.copy_()`, instead of `p.data.copy_()`, work?

H-Huang: IIRC it didn't work and causes a segfault

PR #169, opened 2025-04-22
ProcessGroupNCCL: abort on get_future timeout

d4l3k:
This has some key fixes to improve ProcessGroupNCCL abort behavior when timing out.

This previously had a bug where weren't correctly scheduling the stream_timeout when calling `.get_future()` as called by `allreduce()` in the torchft Manager.

This also modifies `train_ddp.py` to allow for running multiple workers on the same GPU + make it more network heavy so we stress the comm layer and abort.

Test plan:

Added new unit tests to `process_group_test.py`.

```
pytest torchft/process_group_test.py -v -s -x -k test_nccl_timeouts
```

train_ddp + failure injection via `kill`

```
torchx run -- --replicas 16 --max_restarts=10
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -589,15 +589,21 @@ class _WorkCUDATimeout(Work):
         self._timeout = timeout
 
     def wait(self, timeout: Optional[timedelta] = None) -> bool:
-        with self._stream_timeout(self._pg, self._timeout):
-            # not async, just return
-            if self._work is None:
-                return True
+        async_timeout = timeout or self._timeout
+        with self._stream_timeout(self._pg, async_timeout):
+            # In newer versions of PyTorch work may not exist if the call was
+            # not async. In these cases we can just schedule the stream timeout
+            # and return.
+            if self._work is not None:
+                if not self._work.wait():
+                    return False
 
+            # Always use cuda stream for timeout to avoid ProcessGroupNCCL
+            # watchdog firing and crashing the process.
             if timeout is not None:
-                return self._work.wait(timeout)
+                torch.cuda.synchronize()
 
-            return self._work.wait()
+            return True
 
     @classmethod
     @contextmanager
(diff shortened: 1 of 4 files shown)


torchft/process_group.py
```diff
@@ -589,15 +589,21 @@ def __init__(self, pg: ProcessGroup, work: Work, timeout: timedelta) -> None:
...
+            # and return.
+            if self._work is not None:
+                if not self._work.wait():
+                    return False
+
+            # Always use cuda stream for timeout to avoid ProcessGroupNCCL
+            # watchdog firing and crashing the process.
             if timeout is not None:
-                return self._work.wait(timeout)
+                torch.cuda.synchronize()
```
H-Huang, line 604: For my understanding, we are calling `torch.cuda.synchronize()` here so that it will hang and timeout in   `_stream_timeout` which uses `stream_timeout` (https://github.com/pytorch/torchft/blob/main/torchft/process_group.py#L627) to handle the error, is that right? What happens in the case that timeout is `None`?

H-Huang (approved):
LGTM, thanks!

torchft/process_group_test.py
```diff
@@ -553,6 +553,34 @@ def test_nccl_apis(self) -> None:
 
         torch.cuda.synchronize()
 
+    # pyre-fixme[56]: Pyre was not able to infer the type of argument
+    @skipUnless(
+        torch.cuda.is_available() and torch.cuda.nccl.version() >= (2, 25),
+        "needs NCCL >=2.25",
```
fegin, line 559: Do we have any version check in the torchft source code to verify the NCCL version?

fegin (approved):

torchft/process_group.py
d4l3k: In the case that timeout is not None we will instead synchronize in `should_commit`. The stream_timeout logic will still apply and call `abort` if the stream hangs so we'll detect the failure and report it in should_commit.

torchft/process_group_test.py
d4l3k: @fegin we do have some checks (that gate usage of ncclCommAbort) but we don't have any hard checks. I was debating adding a hard dep on PT 2.7+ since 2.6 has many known issues.

Users should only use PT 2.7 RC or nightly

PR #177, opened 2025-04-27
:sparkles: add watch to _TimeoutManager

### LINKED ISSUE #152: Add watchdog to _TimeoutManager+ProcessGroupNCCL to guarantee fast aborts

tushar00jain:
Summary: Fixes #152. Add a thread in _TimeoutManager that periodically schedules work on the event loop that runs callbacks association with timeouts. If the callback hangs, the watchdog exits.

Test Plan: `python -m unittest torchft.futures_test.FuturesTest`

Reviewers: @d4l3k

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchft/futures_test.py
+++ b/torchft/futures_test.py
@@ -1,6 +1,7 @@
 import threading
 from datetime import timedelta
 from unittest import TestCase, skipUnless
+from unittest.mock import Mock, patch
 
 import torch
 from torch.futures import Future
@@ -15,6 +16,13 @@ from torchft.futures import (
 
 
 class FuturesTest(TestCase):
+    def setUp(self) -> None:
+        self._original_watchdog_interval = _TIMEOUT_MANAGER._watchdog_interval
+        _TIMEOUT_MANAGER._watchdog_interval = timedelta(seconds=1)
+
+    def tearDown(self) -> None:
+        _TIMEOUT_MANAGER._watchdog_interval = self._original_watchdog_interval
+
     def test_future_wait(self) -> None:
         # pyre-fixme[29]: Future is not a function
         fut = Future()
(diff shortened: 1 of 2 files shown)


tushar00jain:
The tests are failing for me on trunk as well.

Here's the output from running

```
pytest -vs torchft/process_group_test.py::NormalNcclMultiPgTest
```

[error.log](https://github.com/user-attachments/files/19932240/error.log)

d4l3k (approved):
LGTM, thanks for adding this!

d4l3k:
@tushar00jain those tests are known to be flaky -- working on fixing them now. Usually they pass after rerunning CI

torchft/futures_test.py
```diff
@@ -13,6 +14,8 @@
     stream_timeout,
 )
 
+_TIMEOUT_MANAGER._watchdog_interval = timedelta(seconds=1)
```
H-Huang, line 17: I imagine this modifies the global state and could affect other tests? Will that cause any issues?

H-Huang (approved):
Looks good!

torchft/futures_test.py
d4l3k: That's a good catch -- we should move this into setUp/tearDown to reset it after

PR #179, opened 2025-04-28
ProcessGroupNCCL: timeout ncclCommInit

d4l3k:
Currently we aren't timing out ncclCommInit operations in a safe manner nor timing out if the collective doesn't return immediately.

This has 3 main changes:

* non-blocking mode: This solves ncclCommInit hangs by using non-blocking mode and setting  `TORCH_NCCL_NONBLOCKING_TIMEOUT` if it's not set. The ENV timeout is less than ideal but we don't have a better option in PTD currently.
* wrapping each collective in timeout_context so if NCCL is running in collective blocking mode we can call abort
* adding `set_timeout` to ProcessGroup so we can adjust the timeout in tests to avoid flakiness and reduce failure time for happy path vs failure cases

This fixes the persistent timeout/hang that we've been getting in CI such as https://github.com/pytorch/torchft/actions/runs/14696355293/attempts/4?pr=177

Test plan:

```
pytest -o log_cli=1 torchft/process_group_test.py -v -s -x -k 'NormalNcclMultiPgTest'
```

CI

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -17,7 +17,9 @@ runtime users need to take care to not assume a static rank or world size.
 """
 
 import logging
+import os
 import threading
+import warnings
 from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass
 from datetime import timedelta
@@ -348,6 +350,12 @@ class ProcessGroup(BaseProcessGroup):
         """
         return None
 
+    def set_timeout(self, timeout: timedelta) -> None:
+        """
+        Sets the default timeout for the process group.
+        """
+        raise NotImplementedError("set_timeout not implemented")
+
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}()"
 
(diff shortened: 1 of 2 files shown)


torchft/process_group.py
```diff
@@ -348,6 +350,12 @@ def errored(self) -> Optional[Exception]:
         """
         return None
 
+    def set_timeout(self, timeout: timedelta) -> None:
```
wconstab, line 353: this seems to overshadow the existing `_set_default_timeout` in the Backend interface?  i'd be ok with making `_set_default_timeout` public though.  I just don't think we want to have an alternate way of doing the setting

d4l3k: Yeah, that's intentional for NCCL since NCCL watchdog often will kill the process.

d4l3k: I do agree we should expose set_default_timeout on the PG object directly, current API is subideal and we're depending on private APIs from torchtitan

torchft/process_group.py
```diff
@@ -670,6 +698,15 @@ def __init__(self, timeout: timedelta = timedelta(seconds=60.0)) -> None:
 
         self._errored: Optional[Exception] = None
 
+        NONBLOCKING_TIMEOUT_ENV = "TORCH_NCCL_NONBLOCKING_TIMEOUT"
```
H-Huang, line 701: just confirming, this env variable is read by pytorch before the collective is launched and not when `ProcessGroupNCCL` is initialized right?

H-Huang (approved):
Thanks for fixing those flaky tests!

torchft/process_group.py
d4l3k: Yes, it's read on the first collective

Issue #162, opened 2025-04-15
Example train_ddp.py run command runs only one process, but there are two groups and it is not explain how to use both


rualark:
See here: https://github.com/pytorch/torchft/tree/main?tab=readme-ov-file#example-training-loop-ddp

TORCHFT_LIGHTHOUSE=http://localhost:29510 torchrun --master_port 29501 --nnodes 1 --nproc_per_node 1 train_ddp.py

Two groups by default: https://github.com/pytorch/torchft/blob/main/train_ddp.py#L35

WarrenZhu050413:
Hi Rualark, this is what worked for me. 

After launching the lighthouse, you should launch two groups. Each torchrun group launch instantiates a ManagerServer that can communicate with the lighthouse, and corresponds to a "replica group" in the [design doc](https://docs.google.com/document/d/1OZsOsz34gRDSxYXiKkj4WqcD9x0lP9TcsfBeu_SsOY4/edit?tab=t.0).

```bash
export REPLICA_GROUP_ID=0
export NUM_REPLICA_GROUPS=2

CUDA_VISIBLE_DEVICES=0 TORCHFT_LIGHTHOUSE=http://localhost:29510 torchrun --nnodes=1 --nproc_per_node=1 -- train_ddp.py
```

```bash
export REPLICA_GROUP_ID=1
export NUM_REPLICA_GROUPS=2

CUDA_VISIBLE_DEVICES=1 TORCHFT_LIGHTHOUSE=http://localhost:29510 torchrun --nnodes=1 --nproc_per_node=1 -- train_ddp.py
```

rualark:
Makes sense. Do you want to add this to README.md? Explaining what CUDA_VISIBLE_DEVICES in README.md should also help users?

WarrenZhu050413:
@d4l3k I would be happy to take a stab at this.

d4l3k:
@WarrenZhu050413 feel free! Appreciate any help

We also have torchx support for running locally which does some of this for you. See the .torchxconfig

rualark:
@d4l3k thanks for mentioning .torchxconfig, I did not even notice it before as it is not mentioned in REAMDME.md. Probably makes sense to add it to README.md.

PR #180, opened 2025-04-30
Updated ReadME with HSDP documentation. #162

WarrenZhu050413:
Added instructions for running HSDP using torchFT as per #162. Included how to run using torchX as well as doing so by hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,44 @@ for i in range(1000):
     optimizer.step()
 ```
 
+### Running DDP
+
+After starting the lighthouse server by running:
+
+```sh
+RUST_BACKTRACE=1 torchft_lighthouse --min_replicas 1 --quorum_tick_ms 100 --join_timeout_ms 10000
+```
+
+A test DDP script can be launched with torchX with:
+
+```sh
+torchx run
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -208,6 +208,42 @@ for i in range(1000):
...
 
+### Running HSDP
+
+After starting the lighthouse server by running:
+
+```sh
+RUST_BACKTRACE=1 torchft_lighthouse --min_replicas 1 --quorum_tick_ms 100 --join_timeout_ms 10000
+```
+
+A test HSDP script can be launched with torchX with:
```
d4l3k, line 219: nit: torchx run just runs DDP not HSDP

d4l3k (changes requested):
Thanks for adding this! 

To avoid confusion we should differentiate between fault tolerant HSDP and fault tolerant DDP. The torchtitan example/docs is recommended for how to use HSDP.

README.md
WarrenZhu050413: Thanks for pointing this out. I made the change and added a pointer to torchTitan.

d4l3k (approved):
LGTM

PR #182, opened 2025-05-01
manager: gracefully handle errors from configure+checkpoint

d4l3k:
This makes errors during configure and checkpoint transfer be bubbled up via report_error/errored so we can instead retry rather than crashing.

Test plan:

```
pytest torchft/manager_test.py
```
+ e2e tests


H-Huang (approved):
LGTM

PR #183, opened 2025-05-01
lighthouse: add commit_failures support

d4l3k:
This adds a new field to Lighthouse quorum for `commit_failures`. When a commit failure occurs, a network error has occurred and we thus force a new quorum to ensure that all workers reconfigure.

This is required so #182 will gracefully recover after failure.

Test plan:

```
cargo test
pytest torchft/manager_integ_test.py
```

### DIFF
--- a/src/lighthouse.rs
+++ b/src/lighthouse.rs
@@ -288,6 +288,12 @@ impl Lighthouse {
         if quorum_met.is_some() {
             let participants = quorum_met.unwrap();
 
+            let commit_failure_replica_ids: Vec<String> = participants
+                .iter()
+                .filter(|p| p.commit_failures > 0)
+                .map(|p| p.replica_id.clone())
+                .collect();
+
             // only increment quorum ID if something about the quorum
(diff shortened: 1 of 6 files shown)


src/lighthouse.rs
```diff
@@ -301,6 +301,12 @@ impl Lighthouse {
                     "Detected quorum change, bumping quorum_id to {}",
                     state.quorum_id
                 );
+            } else if participants.iter().map(|p| p.commit_failures).sum::<i64>() > 0 {
+                state.quorum_id += 1;
+                info!(
+                    "Detected commit failures, bumping quorum_id to {}",
```
H-Huang, line 307: nit: maybe not necessary since the manager already logs it, but for lighthouse we could also log the participants with commit failures here?

H-Huang (approved):
Looks good!

src/lighthouse.rs
d4l3k: done

PR #184, opened 2025-05-01
Have RUST_LOG be case-insensitive

H-Huang:
small change so something like `RUST_LOG=WARN` also works as opposed to only `RUST_LOG=warn` working


d4l3k (approved):
LGTM, thanks!

PR #185, opened 2025-05-01
process_group/gloo: support CUDA tensors

d4l3k:
This makes torchft's ProcessGroupGloo correctly support CUDA tensors so we can use Gloo with torchtitan for allreduces.

I also found a bug in Gloo, apparently torchft's test suite is more in depth than the one in core haha https://github.com/pytorch/pytorch/issues/152645

Test plan:

torchtitan on remote cluster

```
pytest torchft/process_group_test.py -k gloo_apis
```


fegin (approved):

H-Huang (approved):
cool! I also wonder if it would be possible to use torchtitan for cpu-only training. That could be cool to show a demo on a single host

Issue #173, opened 2025-04-25
support multiple quorums on the same LighthouseServer


d4l3k:
Having the ability to support multiple quorums on the same lighthouse server would make it much easier to deploy torchft in certain scenarios.

With this feature you could deploy a single lighthouse server and then use it for all jobs running in that cluster by using the job ID. This simplifies discovery and would work with most batch job schedulers.

Two possible designs:

# 1. room_id outside of GRPC

We likely want to make it so you can create a new Lighthouse client with a certain key and it'll automatically isolate the requests to that namespace.

Implementing this cleanly on the server side may be a bit tricky -- we may need to do some manipulation under the hood to instantiate one LighthouseServer instance per incoming request to make this cleaner and avoid polluting the API with "room" ids.

https://github.com/pytorch/torchft/blob/main/src/lighthouse.rs#L601

# 1. room_id as field on heartbeat + quorum methods

This may be simpler in some ways to implement as we can just add a room_id field to all the lighthouse requests and then internally route as necessary. No magic with GRPC services is required.

MattKotzbauer:
Hi! I'd like to try and pick this up using the latter method of including room_id as a field in the heartbeat and quorum methods - would then change the `src/lighthouse.rs`'s state storage from a `Mutex<State>` to a `DashMap<String, Mutex<State>>` to scope state to room. Will try and post a PR by tn 🤙.

Warmly,
Matt

d4l3k:
Put together a diagram for this impl

![Image](https://github.com/user-attachments/assets/89dec294-1ee5-46a3-bc08-926fd791b4ef)

PR #189, opened 2025-05-05
 Support multiple quorums on a single LighthouseServer using gRPC metadata-based room assignment

MattKotzbauer:
GH Issue: https://github.com/pytorch/torchft/issues/173

Extends Lighthouse to to support multiple independent quorums on a single server by tagging each gRPC call with a `room-id` metadata header and feeding requests through a lightweight router that maintains per‐room state with a `DashMap<String, Arc<Lighthouse>>`. `LighthouseClient` now accepts an optional `room_id` argument and automatically injects the corresponding metadata header into each `heartbeat` and `quorum` request, while untagged calls continue to use a default namespace. In `multi_quorum_test.py`, I created 2 clients with distinct room ID's and have them form independent quorums on the same server port.

Open to making any changes to the code or approach 🤙.

Warmly,
Matt

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,10 @@ pub mod lighthouse;
 pub mod manager;
 mod net;
 mod retry;
+mod router;
 mod timeout;
+
+pub use crate::router::Router;
 
 use anyhow::Result;
 use atty::Stream;
@@ -21,6 +24,7 @@ use std::thread::available_parallelism;
 use structopt::StructOpt;
 use tokio::runtime::Runtime;
 use tokio::task::JoinHandle;
+use tokio_stream::wrappers::TcpListenerStream;
 use tonic::transport::Channel;
 use tonic::Status;
 
@@ -33,7 +37,9 @@ pub mod torchftpb {
 }
 
 use crate::torchftpb::lighthouse_service_client::LighthouseServiceClient;
+use crate::torchftpb::lighthouse_service_server::LighthouseServiceServer;
 use crate::torchftpb::manager_service_client::ManagerServiceClient;
+use crate::torchftpb::LighthouseHeartbeatRequest;
 use crate::torchftpb::{
     CheckpointMetadataRequest, LighthouseQuorumRequest, ManagerQuorumRequest, ShouldCommitRequest,
 };
@@ -336,9 +342,12 @@ fn lighthouse_main(py: Python<'_>) -> PyResult<()> {
 }
 
 async fn lighthouse_main_async(opt: lighthouse::LighthouseOpt) -> Result<()> {
-    let lighthouse = lighthouse::Lighthouse::new(opt).await?;
+    let router = Router::new(opt.clone());
 
-    lighthouse.run().await?;
+    tonic::transport::Server::builder()
+        .add_service(LighthouseServiceServer::new(router))
+        .serve(opt.bind.parse::<std::net::SocketAddr>()?)
+        .await?;
 
     Ok(())
 }
@@ -476,13 +485,19 @@ fn convert_quorum(py: Python, q: &torchftpb::Quorum) -> PyResult<Quorum> {
 struct LighthouseClient {
     client: LighthouseServiceClient<Channel>,
     runtime: Runtime,
+    room_id: Option<String>,
 }
 
 #[pymethods]
 impl LighthouseClient {
-    #[pyo3(signature = (addr, connect_timeout))]
+    #[pyo3(signature = (addr, connect_timeout, room_id = None))]
     #[new]
-    fn new(py: Python<'_>, addr: String, connect_timeout: Duration) -> PyResult<Self> {
+    fn new(
+        py: Python<'_>,
+        addr: String,
+        connect_timeout: Duration,
+        room_id: Option<String>,
+    ) -> PyResult<Self> {
         py.allow_threads(move || {
             let runtime = tokio::runtime::Builder::new_multi_thread()
                 .worker_threads(num_threads())
@@ -495,6 +510,7 @@ impl LighthouseClient {
             Ok(Self {
                 client: client,
                 runtime: runtime,
+                room_id: room_id,
             })
         })
     }
@@ -549,6 +565,8 @@ impl LighthouseClient {
                 }),
             });
 
+            let mut request = self.add_room_header(request);
+
             // This timeout is processed on the server side so we also enable
             // keep alives to detect server health.
             request.set_timeout(timeout);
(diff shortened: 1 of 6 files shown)


src/router.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    async fn heartbeat(
+        &self,
+        req: Request<LighthouseHeartbeatRequest>,
+    ) -> Result<Response<LighthouseHeartbeatResponse>, Status> {
+        let id = Self::extract_room_id(req.metadata()).to_owned();
+        let room = self.room(&id).await;
+        <Arc<Lighthouse> as LighthouseService>::heartbeat(&room, req).await
+    }
+}
```
d4l3k, line 88: I think this is fine as is since this is fairly minimal boilerplate per request but I think we can do even better. 

By doing this at the Service layer instead of LighthouseService layer we can have it automatically work for all endpoints on the LighthouseService

Can you look into this and see how feasible it is? If it's not any cleaner we can land this as is

Some pointers: 
* https://docs.rs/tower-service/0.3.3/tower_service/trait.Service.html
* https://github.com/hyperium/tonic/blob/b303caa52ba8bbe8172310be7165a80b7c2a53f8/examples/src/tower/server.rs#L83-L109

There's also https://github.com/teimuraz/tonic-middleware which might be useful

d4l3k (changes requested):
Thanks for putting this together! This looks very promising and excited to have this in torchft.

I think there's a couple of ways we can make this cleaner/more generic that I've commented on

src/lib.rs
```diff
@@ -581,13 +599,29 @@ impl LighthouseClient {
     ) -> Result<(), StatusError> {
         py.allow_threads(move || {
             let mut req = tonic::Request::new(LighthouseHeartbeatRequest { replica_id });
+            let mut req = self.add_room_header(req);
```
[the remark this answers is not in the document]
MattKotzbauer: 👍 Sent as most recent commit

src/router.rs
MattKotzbauer: I tried an initial attempt to do the routing at the Service layer rather than the LighthouseService layer, but have had trouble adapting between the  initial tonic message types (`tonic::Request/Response`) and the Tower message types (`http::Request/Response`) - `tonic::Request/Response` wraps the body in `tonic::body::BoxBody` and carries gRPC-specific extensions, while the Tower stack we’re intercepting expects a bare `http::Request/Response<B>` where the body implements `HttpBody`. I haven't yet found a concise way to do this.

If I were to keep at this, I'd see if I could get something working that relies more on `tonic-middleware` - perhaps there's a way to stay entirely in the `tonic` domain that keeps the implementation and debugging cleaner?

d4l3k: mixing the two is a bit tricky -- we probably need to stay at the tower layer. Why do you need to access the tonic::Request/Response objects? It's all HTTP at the end of the day so seems like we should be able to operate at the tower/http layer and view the metadata as a header?

middleware might work though it may be too high level

MattKotzbauer: Ah I see, it became easier when I had router.rs operate entirely at the tower layer rather than trying to mix Service and tonic. Most recent commit has router.rs at the tower level, which lets us start the lighthouse server with a call to
Server::builder().add_service(router).serve(addr)

src/interceptor.rs~
```diff
@@ -0,0 +1,12 @@
+use tonic::{Request, Status, service::Interceptor};
```
d4l3k, line 1: Is this file intentional?

src/lib.rs
```diff
@@ -654,7 +689,7 @@ impl LighthouseServer {
     /// Returns:
     ///    str: The address of the lighthouse server.
     fn address(&self) -> PyResult<String> {
-        Ok(self.lighthouse.address().to_string())
+        Ok(self.bind.clone())
```
d4l3k, line 692: this unfortunately isn't sufficient -- bind could be something like "0.0.0.0:0" which will bind to a random port. Address needs to be the routable http address i.e. `http://foo.bar:1324`

src/router.rs
```diff
@@ -0,0 +1,110 @@
...
+    torchftpb::lighthouse_service_server::LighthouseServiceServer,
+};
+
+/// Metadata header recognised by both client interceptor and this router.
+pub const ROOM_ID_HEADER: &str = "room-id";
+
+/// gRPC server for a single room (inner state = `Arc<Lighthouse>`).
+type GrpcSvc = LighthouseServiceServer<Arc<Lighthouse>>;
+
+#[derive(Clone)]
```
d4l3k, line 29: why does Router need to be Cloneable?

src/router.rs
```diff
@@ -0,0 +1,110 @@
...
+        rooms: Arc<DashMap<String, Arc<GrpcSvc>>>,
+        tmpl: LighthouseOpt,
+        id: &str,
+    ) -> Arc<GrpcSvc> {
+        if let Some(svc) = rooms.get(id) {
+            return svc.clone();
+        }
+
+        // Build room state once.
+        let lh = Lighthouse::new(tmpl.clone())
```
d4l3k, line 59: can we pass in the `id` into Lighthouse so we can prepend it to the Lighthouse log messages?

torchft/multi_quorum_test.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+
+import datetime as _dt
+import time
+
+import pytest
+
+import torchft._torchft as ext
```
d4l3k, line 8: can we use the torchft.coordination API for this test instead?

torchft/multi_quorum_test.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+
+import datetime as _dt
```
d4l3k, line 3: we usually just do `from datetime import timedelta`

src/router.rs
```diff
@@ -0,0 +1,110 @@
...
+        hdrs.get(ROOM_ID_HEADER)
+            .and_then(|v| v.to_str().ok())
+            .unwrap_or("default")
+    }
+
+    async fn room_service(
+        rooms: Arc<DashMap<String, Arc<GrpcSvc>>>,
+        tmpl: LighthouseOpt,
+        id: &str,
+    ) -> Arc<GrpcSvc> {
```
d4l3k, line 53: Should this be typed `Arc<LighthouseServiceServer>` instead?

src/router.rs
```diff
@@ -0,0 +1,110 @@
...
+        if let Some(svc) = rooms.get(id) {
+            return svc.clone();
+        }
+
+        // Build room state once.
+        let lh = Lighthouse::new(tmpl.clone())
+            .await
+            .expect("failed to create Lighthouse");
+
+        let svc_new = Arc::new(LighthouseServiceServer::new(lh));
```
d4l3k, line 63: Should we be just returning Arc<Lighthouse> from this method and constructing the LighthouseServiceServer wrapper on demand so we don't need to clone it in the parent method?

torchft/multi_quorum_test.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
```
d4l3k, line 1: could move this test to `lighthouse_test.py` to keep it with the rest of the lighthouse tests

d4l3k (changes requested):
@MattKotzbauer this is looking great and thanks for pushing this through! Just need some small cleanups

src/interceptor.rs~
MattKotzbauer: Ah that's unintentional, removing in next commit

src/lib.rs
MattKotzbauer: Hmm, perhaps we could use similar calls as the Lighthouse class uses to resolve host IP and address? Will include a version of this in next commit, though am also down to change it

src/router.rs
MattKotzbauer: I mainly made Router Cloneable so that calls to tonic's [add_service](https://docs.rs/tonic/latest/tonic/transport/struct.Server.html#method.add_service) would compile when constructing the LighthouseServer in src/bin/lighthouse.rs and src/lib.rs

src/router.rs
MattKotzbauer: Sounds good, will include in next commit

torchft/multi_quorum_test.py
MattKotzbauer: Sg, including in next commit (moving test into lighthouse_test.py and will use existing imports)

torchft/multi_quorum_test.py
MattKotzbauer: Sg, including in next commit

src/router.rs
MattKotzbauer: Oh true, will include in next commit

src/router.rs
MattKotzbauer: This sounds good - am changing in line with the below thread (returning an Arc<Lighthouse> and then wrapping with a LighthouseServiceServer once the method returns).

PR #190, opened 2025-05-06
Add traceback to manager reported error

H-Huang:
Before:
image

After:
image

### DIFF
--- a/torchft/manager_test.py
+++ b/torchft/manager_test.py
@@ -444,7 +444,8 @@ class TestManager(TestCase):
         manager._pg.errored.return_value = injected_failure
 
         self.assertFalse(manager.should_commit())
-        self.assertEqual(manager._errored, injected_failure)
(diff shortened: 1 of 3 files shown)


torchft/manager_test.py
```diff
@@ -444,7 +444,7 @@ def test_pg_errored(self, client_mock: MagicMock) -> None:
         manager._pg.errored.return_value = injected_failure
 
         self.assertFalse(manager.should_commit())
-        self.assertEqual(manager._errored, injected_failure)
+        self.assertIsInstance(manager._errored, Exception)
```
d4l3k, line 447: can we check .original_exception here instead? and for all the tests below including raise?

The type of error I think is pretty important for some of these tests to distinguish different code branches

d4l3k (changes requested):

d4l3k (approved):
LGTM

PR #191, opened 2025-05-08
Fix ManagedProcessGroup when used with c10d APIs

H-Huang:
Fixing this here instead of the changes in https://github.com/pytorch/torchtitan/pull/1170

Tested this change in Torchtitan


d4l3k (approved):
LGTM

gravityknife:
Awesome!

PR #187, opened 2025-05-05
Distinguishing between replica rank and group rank across the project (#181)

WarrenZhu050413:
Distinguishes between replica rank and group rank across the projects as per #181.

Kept the API the same for all except for the DistributedSampler. I found the terminology especially confusing there. And since it is not used in torchTitan I changed the API.

All tests have passed. However, I am getting the following linter error. It doesn't seem like a problem with my code. 

```sh
@warren# lintrunner -a
Warning: Could not find a lintrunner config at: '.lintrunner.private.toml'. Continuing without using configuration file.

>>> General linter failure:

  Advice (pyre) command-failed
    Failed due to JSONDecodeError:
    Expecting value: line 1 column 1 (char 0)
Successfully applied all patches.
(/srv/apps/danny/miniconda3/envs/warren/torchtitan) root@sz-k8s-master:/srv/apps/warren/fork/torchft# 
```

# Changes

## Globally
Changed recovery_src_rank, recovery_dst_ranks, max_rank to `_replica_rank`. 

## In Manager.py:

```python
rank -> group_rank
self._rank -> self._group_rank
self._participating_rank -> self._participating_replica_rank
self._participating_world_size -> self._participating_replica_world_size 
max_world_size -> max_replica_world_size world_size -> group_world_size 
self._world_size_mode -> self._replica_world_size_mode
```

## In Data.py

Changed API along with internal reference:

```python
rank -> group_rank
replica_group -> replica_rank
```

## In Rust Code

`ShouldCommitRequest:` rank -> group_rank
`self._client._quorum`: rank -> group_rank
Did not change CheckpointMetadataRequest's proto interface since it should be it seems less tied to the specific group_rank/replica rank distinction made here.

Issue #181, opened 2025-04-30
Unify replica terminology across the project


rualark:
Currently in different places of code and docs these things are named differently:
- replica groups are sometimes called replicas (and replica_id sometimes means replica_group_id, replica_rank means replica_group_rank)
- replicas sometimes mean workers, but sometimes they mean replica groups
- etc.

Examples of the terminology problem:
https://github.com/pytorch/torchft/blob/abc660b6a60b015dbff2a591508d1b4932ab20fa/torchft/torchx.py#L64
https://github.com/pytorch/torchft/blob/abc660b6a60b015dbff2a591508d1b4932ab20fa/torchft/torchx.py#L69
https://github.com/pytorch/torchft/blob/abc660b6a60b015dbff2a591508d1b4932ab20fa/src/manager.rs#L391

What can be done:
1. define unified terminology for the key objects
2. update docs and code to be aligned

WarrenZhu050413:
Another related confusion that I had first encountering the code base is whether `rank` referred to the rank of the replicas or the rank of the devices within a replica group.

E.g. `max_rank` and `primary_rank` in `manager.rs` seems to refer to the `replica_rank`, whilst `Manager._rank` refers to the rank within a replica group.

d4l3k:
I've tried to be consistent but yes, it is a bit confusing having two rank/replicas concepts.

Unfortunately for the code in `torchx.py` it's interacting with torchelastic which doesn't have the concept of replica groups so we don't have much choice other than do the translation there.

For the code in manager we should really distinguish everywhere between group_rank and replica_rank

rualark:
Maybe torchft does not need a concept of replica - only replica groups. Then "replica" term will be used only in torchx. This will be simpler if the same term like "replica" does not mean two different things in the same file.

rualark:
As replica group is actually a collection of nodes that have the full model and process a shard of data, should it be named data group or ddp group or data shard or ddp shard?

WarrenZhu050413:
@rualark I believe that replica_group is a commonly used term in distributed systems. It may be good to keep the terminology consistent with the literature.

WarrenZhu050413:
@d4l3k  I would also be happy to make amendments to make everything group_rank/replica_rank.

I searched through all the occurances of "rank" in the torchFT repo, and came up with the required changes. However, I am not sure whether the names in the exposed methods should change. 

E.g. 
```python
manager.participating_rank() -> manager.participating_replica_rank()
```

[...]

PR #193, opened 2025-05-13
Fix compatibility between data.py and train_ddp.py with replica rank terminology (#187)

WarrenZhu050413:
Commit `652a009 Distinguishing between replica rank and group rank across the project (#181) (#187)` broke the `train_ddp.py` file. This commit fixes it and changes the terminology in `data.py` to use the more accurate terminology of `replica_group_id` rather than `replica_rank`.


d4l3k (approved):
LGTM

PR #197, opened 2025-05-20
Fix DiLoCo with DTensor

H-Huang:
Hitting an error with `RuntimeError: The size of tensor a (128256) must match the size of tensor b (128220) at non-singleton dimension 0` when restoring the parameters in DiLoCo.

TODO: add tests and repro this locally


torchft/local_sgd.py
```diff
@@ -249,7 +249,7 @@ def _restore_parameters(self) -> None:
                     # we averaged the local version of the tensor so need to copy it back as a DTensor
                     p.data.copy_(
                         DTensor.from_local(
-                            self.original_parameters[name], p.device_mesh, p.placements
+                            self.original_parameters[name], p.device_mesh, p.placements, shape=p.shape, stride=p.stride()
```
fegin, line 252: I'm curious why shape changes?

H-Huang: For 10 ranks, the original parameter was shape [128256, 4096]

After converting to dtensor each rank had a local tensor of [12826, 4096], except for rank 9 which was [12822, 4096]

I saw this note in the docs about needing to add shape arg for [non-even tensors](https://docs.pytorch.org/docs/stable/distributed.tensor.html#:~:text=Note%20this%20needs%20to%20be%20provided%20if%20the%20shape%20of%20local_tensor%20are%20different%20across%20the%20ranks.). Though it is strange because I cannot repro this in a test yet. However this does fix the issue when running on the cluster

fegin (approved):

PR #196, opened 2025-05-20
Added proactive heartbeat timeout failure propagation (#164) (#188)

WarrenZhu050413:
# Overview

This PR improves failure detection speed of torchFT through proactive failure recovery. The Manager now listens to Lighthouse failure notifications and aborts hanging collectives immediately instead of waiting for NCCL/Gloo time-outs.

# Basic demonstration

You can experiment with proactive failure recovery mode by:

```sh
export TORCHFT_PROACTIVE_RECOVERY=1
```

With this enabled, the manager will listen to the Lighthouse server for heartbeat failures of other replica groups and break from a hanging allreduce.

You can test this out by running `train_ddp_proactive.py`

On shell 1 (one replica groups starts initial training):
```sh
export REPLICA_GROUP_ID=0
export NUM_REPLICA_GROUPS=2
export TORCHFT_PROACTIVE_RECOVERY=1

CUDA_VISIBLE_DEVICES=0 TORCHFT_LIGHTHOUSE=http://localhost:29510 torchrun --master_port=29600 --nnodes=1 --nproc_per_node=1 -- train_ddp_proactive.py
```

On shell 2 (a second replica group joins):
```sh
export REPLICA_GROUP_ID=1
export NUM_REPLICA_GROUPS=2
export TORCHFT_PROACTIVE_RECOVERY=1

CUDA_VISIBLE_DEVICES=1 TORCHFT_LIGHTHOUSE=http://localhost:29510 torchrun --master_port=29601 --nnodes=1 --nproc_per_node=1 -- train_ddp_proactive.py
```

You should observe that the process with replica group id 1 will exit early, and the process with replica group id 0 will quickly resume training. If the same script is ran with after setting `export TORCHFT_PROACTIVE_RECOVERY=0`, you should observe that the process with replica group id 1 will hang for dozens of seconds before continuing.
```sh
INFO:torchft.manager:[train_ddp_0:81a52ce4-d803-4f22-a0c3-54f3b4a88c89/0 - step 10] Setting error processor thread stop event
INFO:torchft.manager:[train_ddp_0:81a52ce4-d803-4f22-a0c3-54f3b4a88c89/0 - step 10] Waiting for error processor thread to complete
INFO:torchft.manager:[train_ddp_0:81a52ce4-d803-4f22-a0c3-54f3b4a88c89/0 - step 10] Error processor thread shutdown completed.
[...]

Issue #130, opened 2025-03-11
Add config sharing from Lighthouse


d4l3k:
It would be nice to be able to broadcast configs from lighthouse to all workers

We can add a new `config` endpoint to Lighthouse/Manager that workers can call to fetch it

WarrenZhu050413:
If more granular failure information (e.g. which client has failed) is accessible to the lighthouse, this could enable dynamic reconfiguration within Model Parallel groups (e.g. [Oobleck](https://arxiv.org/pdf/2309.08125), [Varuna](https://arxiv.org/pdf/2111.04007)), extending FT beyond per-replica group granularity.

PR #202, opened 2025-05-24
Add config sharing from Lighthouse with UI support (#130)

WarrenZhu050413:
## Add config sharing from Lighthouse with UI support (#130)

Implemented GetConfig RPC for the Lighthouse

### Summary

Implemented and tested the GetConfig RPC functionality for the Lighthouse service, enabling configuration broadcasting from the lighthouse to client applications. Along with this, added UI support 

### Tests

- Added GetConfigRPC to the Lighthouse
- Added Python bindings to the RPC
- Added `test_get_config_rpc()` in `torchft/lighthouse_test.py` to test that the GetConfig RPC endpoint to ensure it returns configuration data

### Current Implementation Details

The GetConfig RPC currently:
- Returns configuration data as `HashMap<String, String>` (Python `dict[str, str]`)
- Supports timeout configuration (default: 5 seconds)
- Loads configuration from JSON files specified via `--lighthouse-config` parameter
- Returns empty configuration when no config file is provided
- Handles invalid JSON by returning empty configuration

### Discussion: Incorporating config sharing into the entire torchFT workflow

I am personally really excited about incorporating config sharing into torchFT, as I believe it enables a lot of intereting possibilities.

With config sharing, the lighthouse could reconfigure the training process at runtime according to user input/information that it has, acting as an orchestrator for the training process.

Some usecases that I can think of include:
a) **Dynamic Training Algorithm Configuration**
   - Adjust DiLoCo synchronization frequency based on network conditions/loss curve convergence (this could be in the examples to showcase the config reconfiguration feature)
   - Change the optimizer that the training process uses at runtime
   - Modify learning rates based on training progress (rather than relying on LR scheduler)
   - Enable interactive batch size adjustments

b) **Advanced Fault Tolerance**
   - Coordinate parallelism structure changes upon worker failures
   - Reroute messages in MoE models when nodes fail
   - Rerouting to healthy workers in pipeline parallelism upon a worker failure
   - Have a time series model that predicts worker failures and route to that worker upon failure.

With these in mind, I want feedback on two questions:

1. Whether the usecases that I envision above belong to torchFT, or some other newer, experimental framework
2. Taking into account these factors, what is the best way for the `Manager` to use the config sharing functionality?

#### **Approach 1: Request-Response via Manager**

The most torchFT idiomatic way to use the config sharing functionality is to have a two-stage RPC: Worker -> Manager -> Lighthouse -> Manager -> Worker. At the beginning of each training iteration, ManagerClient can call a `GetConfig` RPC to the manager, and then send the config to the worker.

However, this approach has two limitations:

1. Because all the communication between the worker and the Lighthouse go through a ManagerServer, the Lighthouse does not have direct access to the Worker. In the configuration context, it is natural for every client under a ManagerServer to receive the same configuration.
2. Relying on Request-Response also decreases the speed at which the configs are broadcasted to the workers. This would require the worker to request and wait for a response for the Lighthouse (of course, this could be done through a prior asynchronous call to avoid blocking on the config, but the relevancy of the received  message may still be low).
   1. A streaming approach could be adopted here. However, because the streaming code is itself relatively complex, managing a two-stage streaming handoff process increases a lot of complexity into the code and makes it hard to maintain, as many logic has to be written twice.

### **Approach 2: Direct Streaming from the Lighthouse to the workers**

Because of the above considerations, I am considering extending the experimental `FailureStream` implemented in [PR 196](https://github.com/pytorch/torchft/pull/196) to become a generic `LighthouseStream` that can handle multiple message types.

This addresses the two limitations of the Request-Response RPC code.

However, it has its own problems:

a) It breaks the Worker -> Manager -> Lighthouse structure that was previously maintained
b) It potentially decreases the scalability of torchFT, since the Lighthouse needs to maintain connections to all the workers


WarrenZhu050413:
config UI

^UI

PR #205, opened 2025-05-30
Generalize DiLoCo to support Streaming

tushar00jain:
Summary:
- Add option to perform qunatized allreduce in torchft manager
- Update user level API's for DiLoCo to also support Streaming DiLoCo -- it takes a list of modules now as input
- Create a class `_StreamingDiLoCoFragment` used by DiLoCo to support streaming. Each fragment independently determines its schedule (when to send/sync).
- Adding support for "alpha" and "tao" parameters from the paper are left as a TODO. Plan to add this in a separate PR.

Test Plan:
```
$ pytest -vs torchft/local_sgd_integ_test.py
$ pytest -vs torchft/local_sgd_test.py
```


dzmitry-huba (approved):

torchft/manager.py
```diff
@@ -267,7 +291,9 @@ def shutdown(self, wait: bool = True) -> None:
             self._manager.shutdown()
         self._executor.shutdown(wait=wait)
 
-    def allreduce(self, tensor: torch.Tensor) -> torch.futures.Future[torch.Tensor]:
+    def allreduce(
+        self, tensor: torch.Tensor, should_quantize: bool = False
```
H-Huang, line 281: I would also update some of the tests in `manager_test.py` to also include using the `should_quantize` flag. Can be done in a follow up PR

torchft/manager.py
```diff
@@ -47,6 +48,29 @@
...
 
+try:
+    # pyre-ignore[21]: Could not find a module corresponding to import `triton`
+    import triton
+
+    from torchft.collectives import allreduce_quantized
+except ImportError:
+    from torch import cuda
+
+    def allreduce_quantized(
```
H-Huang: nit: is this stub necessary? Can't we just have a constant like `TRITON_AVAILABLE` and then check that in the if statement in the implementation

torchft/local_sgd.py
```diff
@@ -400,8 +351,148 @@ def _allreduce_bucketized(self) -> None:
...
+    parameters will reset back to the last time DiLoCo synchronized.
+
+    DiLoCo paper: https://arxiv.org/pdf/2311.08105
+    Streaming DiLoCo paper: https://arxiv.org/pdf/2501.18512
+    """
+
+    def __init__(
+        self,
+        manager: Manager,
+        model_fragments: List[nn.Module],
```
H-Huang, line 383: nit: maybe from an API / UX perspective we can support `nn.Module | List[nn.Module]` with the specification that passing in a single nn.Module means whole model.

torchft/local_sgd.py
```diff
@@ -400,8 +351,148 @@ def _allreduce_bucketized(self) -> None:
...
+
+        self._local_optimizer = inner_optimizer
+
+        self._fragments: List[_StreamingDiLoCoFragment] = [
+            _StreamingDiLoCoFragment(
+                manager,
+                model_fragment,
+                math.floor((sync_every / len(model_fragments)) * (i + 1)),
+                inner_optimizer,
+                # TODO: Support different outer optimizers for each fragment
```
H-Huang, line 447: oh interesting, is that mentioned in the paper?

tushar00jain: I think they should be different otherwise things like momentum end up being the same for all fragments? Maybe it's not very important though

torchft/local_sgd.py
tushar00jain: Think it's better to have a smaller api surface and avoid having a special case?

torchft/manager.py
tushar00jain: Less configuration options 🥲 let's say someone changes platforms, then they can start using triton automatically without having to modify the constant. Also avoids having us to configure CI properly for different platforms.

PR #206, opened 2025-06-03
Add recovery for streaming diloco

tushar00jain:
The recovery logic for new nodes joining the quorum is different compared to regular diloco. Primarily because we need to make sure all nodes allreduce fragments in the same order otherwise we can deadlock. In diloco, there's only 1 fragment so there's nothin to order.

- Decouple scheduling and waiting for allreduce futures in `_StreamingDiLoCoFragment`
- Parallelize the allreduce's needed by `bucketize_and_allreduce`
- Move the decision making logic about scheduling and blocking for allreduce to `DiLoCo` instead of doing that independently in each fragment. So that we can the recovery logic in one place. The step counts also need to be centralized.  See comments in `DiLoCo::_step_post_hook`
- Retry `quorum` indefinitely
- Move `local_step` to `manager` -- added comments on why this is needed in `DiLoCo::_step_post_hook`

Test Plan:

```
$ pytest -vs torchft/local_sgd_integ_test.py
$ pytest -vs torchft/local_sgd_test.py
```

Streaming DiLoCo needs one more change to get working, until then it's disabled through asserts. We have some existing tests for DiLoCo so banking on those. The behavior also doesn't change much for DiLoCo, 

- Except quorum is done after commit instead of before the starting allreduce
- Even if `local_step` is different on different nodes, since there is only one fragment, they all allreduce together so deadlocks are not possible

Plan to add Streaming DiLoCo related tests in a separate diff. 

PS: Had to disable diloco integration test on osx - the test is probably hanging when destroying the gloo process group.

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -10,6 +10,7 @@ This module implements a fault tolerant version of LocalSGD and related methods.
 """
 import logging
 import math
+import threading
 from types import TracebackType
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type
 
@@ -176,18 +177,29 @@ class _StreamingDiLoCoFragment:
         use_bucketization: bool = False,
         bucket_cap_mb: Optional[int] = None,
         should_quantize: bool = False,
+        fragment_sync_delay: int = 0,
+        fragment_update_alpha: float = 0.0,
     ) -> None:
+        if fragment_sync_offset > sync_every:
+            raise ValueError("Fragment must be synced once before `sync_every` steps")
+
         self._manager = manager
         self._model_fragment = model_fragment
         self._fragment_sync_offset = fragment_sync_offset
         self._local_optimizer = inner_optimizer
-        self._local_step = 0
         self._sync_every = sync_every
         assert sync_every >= 1, "sync_every must be greater than or equal to 1"
         self._backup_device = backup_device
         self._pin_memory = pin_memory
+        self._fragment_sync_delay = fragment_sync_delay
+        self._fragment_update_alpha = fragment_update_alpha
 
         self._outer_optimizer = outer_optimizer
+
+        # Stores pending all reduce
+        self._allreduce_futures: list[
+            torch.futures.Future[None] | torch.futures.Future[torch.Tensor]
+        ] = []
 
         if bucket_cap_mb is not None:
             self.bucket_cap_mb = int(bucket_cap_mb * 1024 * 1024)
@@ -236,22 +248,34 @@ class _StreamingDiLoCoFragment:
                 else:
                     p.data.copy_(self.original_parameters[name], non_blocking=False)
 
-    def sync(self) -> None:
+    def wait(self) -> None:
         """
-        Synchronizes and averages the model weights across the manager.
+        Waits for the previously scheduled allreduce to finish
         """
-        self._local_step += 1
 
-        if (self._local_step - self._fragment_sync_offset) % self._sync_every != 0:
-            return
+        for work in self._allreduce_futures:
+            work.wait()
 
-        self._manager.start_quorum()
-        self._perform_sync()
+        self._allreduce_futures = []
 
-    def _perform_sync(self) -> None:
+    def should_prepare_fragment(self, step: int) -> bool:
         """
-        Overrides the sync method to calculate the pseugradient, average them across the manager group, and
-        step using the outer optimizer.
+        Determines if the fragment should be asynchronously sent to other replicas
+        """
+        step_to_prepare = step - self._fragment_sync_offset
+        return step_to_prepare % self._sync_every == 0
+
+    def should_sync_fragment(self, step: int) -> bool:
+        """
+        Determines if the fragment should be synchronized with other replicas
+        """
+        step_to_sync = step - self._fragment_sync_offset - self._fragment_sync_delay
+        return step_to_sync % self._sync_every == 0
+
+    def prepare_sync(self) -> None:
+        """
+        Calculate the pseugradient, average them across the manager group and starts
+        allreduce on the pseudo-gradients but doesn't wait for it to finish.
         """
         # Set the .grad field of each parameter to its pseudogradient
         for name, p in self._model_fragment.named_parameters():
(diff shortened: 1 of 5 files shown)


torchft/manager.py
```diff
@@ -763,6 +769,36 @@ def current_step(self) -> int:
         """
         return self._step
 
+    def set_local_step(self, step: int) -> None:
```
d4l3k: thoughts on doing this vs adding a way to register an additional `load_state_dict/state_dict` function for libraries?

Could make `set_state_dict_fns` more generic and have a `register_state_dict_fns` that could be used to register the DiLoCo wrapper which can then keep local_step internal instead of needing all these step manipulation functions

torchft/local_sgd.py
```diff
@@ -176,19 +176,27 @@ def __init__(
...
         assert sync_every >= 1, "sync_every must be greater than or equal to 1"
         self._backup_device = backup_device
         self._pin_memory = pin_memory
+        self._fragment_sync_delay = fragment_sync_delay
+        self._fragment_update_alpha = fragment_update_alpha
 
         self._outer_optimizer = outer_optimizer
 
+        # Stores pending all reduce
+        self._allreduce_futures: List[
```
d4l3k: nit: can just use `list[...]

torchft/local_sgd.py
```diff
@@ -325,9 +357,12 @@ def bucketize_and_allreduce(
...
                 bucket_size_bytes // tensors[0].element_size(), total_size - offset
             )
-            flat_buffer = torch.zeros(chunk_size, dtype=dtype, device=device)
+            flat_buffer: torch.Tensor = torch.zeros(
+                chunk_size, dtype=dtype, device=device
+            )
 
-            pack_offset, bucket_tensors = 0, []
+            pack_offset: int = 0
+            bucket_tensors: List[Tuple[torch.Tensor, int, int]] = []
```
d4l3k: nit: `list[tuple[...]]`

torchft/local_sgd_integ_test.py
```diff
@@ -139,7 +139,7 @@ def state_dict() -> Dict[str, Dict[str, object]]:  # pyre-ignore[53]
         if device.type == "cuda":
             pg = ProcessGroupBabyNCCL()
         else:
-            pg = ProcessGroupGloo()
+            pg = ProcessGroupGloo(timeout=timedelta(seconds=10))
```
d4l3k, line 143: nice catch

torchft/local_sgd_integ_test.py
```diff
@@ -139,7 +139,7 @@ def state_dict() -> Dict[str, Dict[str, object]]:  # pyre-ignore[53]
...
         manager = Manager(
             pg=pg,
             min_replica_size=2,
```
d4l3k, line 146: do we have any tests verifying local step is updated on recovery?

d4l3k (approved):
changes look reasonable

torchft/manager.py
tushar00jain: that sounds like a good idea, specially if we add more things to the state

torchft/local_sgd.py
```diff
@@ -176,19 +176,27 @@ def __init__(
...
         self._manager = manager
         self._model_fragment = model_fragment
         self._fragment_sync_offset = fragment_sync_offset
         self._local_optimizer = inner_optimizer
-        self._local_step = 0
         self._sync_every = sync_every
         assert sync_every >= 1, "sync_every must be greater than or equal to 1"
         self._backup_device = backup_device
         self._pin_memory = pin_memory
+        self._fragment_sync_delay = fragment_sync_delay
```
H-Huang, line 194: probably need some validation logic to check relationships between  `fragment_sync_delay` , `fragment_sync_offset` and `sync_every`

e.g. `fragment_sync_delay` < `fragment_sync_offset`, `fragment_sync_offset` < `sync_every`, etc.

torchft/local_sgd.py
```diff
@@ -488,11 +529,84 @@ def __exit__(
 
         return False  # Propagate exceptions
 
+    def _wait(self) -> None:
+        """
+        Waits for allreduce to finish on all fragments
+        """
+        for fragment in self._fragments:
```
H-Huang, line 569: how come we call `wait()` for all fragments?

torchft/local_sgd.py
```diff
@@ -488,11 +529,84 @@ def __exit__(
...
+                #
+                # In the happy path, we don't need to reeset the step because --
+                # Nodes participating in the commit bumped their `max_step`.
+                # Any new nodes will take `local_step` from one of these nodes, which must
+                # be the same across all nodes because they took the same number of steps
+                # since the last commit to get to the most recent commit.
+
+                self._manager.set_local_step(0)
+
+            self._quorum_loop()
```
H-Huang, line 657: why do we do the loop as the end as well?

torchft/local_sgd.py
tushar00jain: after commit failure, we want to start from a clean slate. this also means setting the allreduce future's on the fragments to null. probably not needed since we only call this when all allreduce's failed anyway. but just being consistent to mimic starting from a clean slate.

torchft/local_sgd.py
tushar00jain: we want to avoid doing allreduce after quorum failed. maybe a different quorum formed without this node, so this node will incorrectly try to allreduce potentially on an incorrect fragment because the `local_step` is also out of sync. the replica will need recovery later anyway. so in case it didn't crash (network errors), we can save some training data by looping here, otherwise that training data goes to waste after recovery

torchft/local_sgd_integ_test.py
tushar00jain: was planning to add tests in a separate diff without blocking implementation review... streaming diloco correctness will need a bunch. i can update this diff with some small tests though

H-Huang (approved):
LGTM!

dzmitry-huba (approved):

torchft/manager.py
tushar00jain: updated this, want to take a look again?

torchft/local_sgd_integ_test.py
tushar00jain: we're using standard manager functions now so 

- manager test validates we can set the state dict
- http transport tests validate state dict is properly transferred

torchft/manager.py
```diff
@@ -262,11 +266,18 @@ def __init__(
         self._participating_replica_rank: Optional[int] = None
         self._participating_replica_world_size: int = 0
 
-    def set_state_dict_fns(
```
d4l3k, line 265: this is used by torchtitan so we need to keep this for backwards compatibility

d4l3k (commented):
looks good, thanks for updating the state_dict bits -- we do need to avoid breaking torchtitan though with set_state_dict_fns

PR #207, opened 2025-06-03
Introduce FP8 reduce-scatter along with SUM and more base types

dzmitry-huba:
This change introduces several extensions to the quantized collectives:

- FP8 reduce-scatter
- float16 and bfloat16 base types for the input tensors
- SUM reduce operation support

### DIFF
--- a/torchft/collectives.py
+++ b/torchft/collectives.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import math
+
 from typing import TYPE_CHECKING
 
 import torch
@@ -11,11 +13,12 @@ import torch
 # pyre-ignore[21]: Could not find a module corresponding to import `triton`
 import triton
 from torch import cuda
+from torch.distributed import ReduceOp
 from torch.distributed.distributed_c10d import (
     AllgatherOptions,
     AllreduceOptions,
     AllToAllOptions,
-    ReduceOp,
+    ReduceScatterOptions,
 )
 from torch.futures import Future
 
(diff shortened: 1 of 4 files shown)


torchft/collectives_test.py
```diff
@@ -24,7 +24,24 @@
...
 else:
-    from torchft.collectives import allreduce_quantized
+    from torchft.collectives import (
+        allocate_reduce_scatter_output,
+        allreduce_quantized,
+        get_padded_sizes,
+        reduce_scatter_quantized,
+    )
+
+    def _check_result_tolerance(
```
H-Huang, line 34: could consider using [torch.allclose()](https://docs.pytorch.org/docs/main/generated/torch.allclose.html)

dzmitry-huba: Unfortunately allclose will not work here. Each of the elements in the tensor may have different magnitude of the difference due to quantization. With allclose I will have to relax the constraints more to accommodate it. Instead here I'm computing an average difference across all elements and ensure that on average the result is within tolerance (0.05 means within 5%)

torchft/quantization.py
```diff
@@ -16,28 +16,35 @@
...
 SCALE_TL_DTYPE_BYTES = tl.constexpr(4)
-TL_MAX_FP8 = tl.constexpr(448.0)
 
 BLOCK_SIZE_T: int = 2048
 
 
 # pyre-ignore[11]: Annotation `tl.constexpr` is not defined
+def _get_fp8_max() -> tl.constexpr:
+    if cuda.get_device_capability() >= (9, 0):
+        return tl.constexpr(448.0)
```
H-Huang, line 32: thats interesting 👀 TIL

torchft/collectives_test.py
```diff
@@ -46,21 +63,22 @@ def _run_parallel_collectives(
...
             tolerance: float,
+            reduce_op: ReduceOp,
+            dtype: torch.dtype,
         ) -> None:
             cuda.set_device(device)
             inp = (
                 torch.rand(
                     tensors_num * tensor_size,
-                    dtype=torch.float32,
+                    dtype=dtype,
```
H-Huang, line 81: curious, if we run with dtypes that aren't supported, is it silently wrong? Or do things blow up

torchft/collectives.py
```diff
@@ -29,18 +31,254 @@
...
+            quantized_reduce_scatter,
+            1,
+        )
+
+        # pyre-ignore[29]
+        return _QuantizedOpFuture(
+            sync_stream,
+            [
+                quantized_inputs,
+                quantized_inputs_out,
```
H-Huang, line 277: do we do any "unpadding" for the output tensors if the input get padded?

torchft/collectives.py
```diff
@@ -29,18 +31,254 @@
...
+        # Allocate output tensor where all-reduce results will be stored
+        quantized_inputs_out = torch.zeros_like(quantized_inputs)
+        # Collect chunks and their scales from other ranks
+        process_group.alltoall_base(
+            quantized_inputs_out.view(world_size, -1),
+            quantized_inputs.view(world_size, -1),
+            [],
+            [],
+            _to_alltoall_options(reducescatter_opts),
+        ).wait()
```
tushar00jain, line 247: we want all this to happen while we continue training. is it a common pattern to call `wait` inside a stream? does it block the cpu/gpu?

tushar00jain: also is the quantization only supported for nccl? should this also be supported for gloo?

PR #209, opened 2025-06-04
Group work for each commit

tushar00jain:
# Summary

## Context

- `should_commit` currently waits for all scheduled work but we need to only wait for work for a fragment
- otherwise, we need to wait for all fragments, which just makes it regular diloco

## Implementation

- removed cuda synchronize when we check errors on process group otherwise we also wait for all the work to finish on the process group
- **_instead we synchronize on the current stream but this might not be needed though because the current steam is doing training only -- @d4l3k? also baby pg doesn't override `errored()` method_**
- remove waiting on pending futures in the manager, the callers need to ensure now that they wait for the futures to complete
- had to change manager `allreduce` to support using custom streams and sprinkle event to this stream when future is ready
  - this way the users of the api can synchronize on the custom stream to make the cpu wait for the work to finish
  - it avoids exposing `Work` to users
  - this also works for baby nccl, currently it was just syncing on the main stream to wait for the allreduce work to finish

## TODO's

- removed support from quantization for now, will add it back once i make sure it works with all pg's
- unit/integration tests for streaming diloco
- tests to make sure the sync logic works as intended with all pg's

# Test Plan

- added a script `train_diloco.py` and gathered some profiles to make sure fragment send/sync happens at the right steps (using nccl)
- compare streaming vs non-streaming
- streaming performs better

## Streaming

- 2 fragments, H=10, T=5
- fragments get sent and synced at every 5 steps
- **_there's still a problem -- if there's already an allreduce in flight, a wait will block the allreduce that's issued on the current step as well. likely because there being 1 stream for networking. we need to create a separate stream for each of these allreduce to fix this. ideally nccl pg can offer an api to specify which stream users want to use_**

image

## Non-Streaming

- 1 fragment, H=10, T=0
- fragments get sent and synced (in the same step) at every 10 steps

image

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -11,6 +11,7 @@ This module implements a fault tolerant version of LocalSGD and related methods.
 import logging
 import math
 import threading
+from contextlib import nullcontext
 from types import TracebackType
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type
 
@@ -197,9 +198,10 @@ class _StreamingDiLoCoFragment:
         self._outer_optimizer = outer_optimizer
 
         # Stores pending all reduce
-        self._allreduce_futures: list[
-            torch.futures.Future[None] | torch.futures.Future[torch.Tensor]
-        ] = []
+        self._allreduce_futures: list[torch.futures.Future[torch.Tensor]] = []
+        self._stream: Optional[torch.cuda.Stream] = (
+            torch.cuda.Stream() if torch.cuda.is_available() else None
+        )
 
         if bucket_cap_mb is not None:
             self.bucket_cap_mb = int(bucket_cap_mb * 1024 * 1024)
@@ -222,6 +224,7 @@ class _StreamingDiLoCoFragment:
                 t = t.pin_memory()
             self.original_parameters[name] = t
 
+    @torch.profiler.record_function("torchft::local_sgd::save_parameters")
     def save_parameters(self) -> None:
         with torch.no_grad():
             # TODO: consider running copy on a separate stream
@@ -229,6 +232,7 @@ class _StreamingDiLoCoFragment:
                 param_to_local = extract_local_tensor(p.data)
                 self.original_parameters[name].copy_(param_to_local, non_blocking=True)
 
+    @torch.profiler.record_function("torchft::local_sgd::restore_parameters")
     def restore_parameters(self) -> None:
         with torch.no_grad():
             # TODO: consider running copy on a separate stream
@@ -248,6 +252,7 @@ class _StreamingDiLoCoFragment:
                 else:
                     p.data.copy_(self.original_parameters[name], non_blocking=False)
 
+    @torch.profiler.record_function("torchft::local_sgd::wait")
     def wait(self) -> None:
         """
         Waits for the previously scheduled allreduce to finish
@@ -255,6 +260,9 @@ class _StreamingDiLoCoFragment:
 
         for work in self._allreduce_futures:
             work.wait()
+
+        if self._stream is not None:
+            self._stream.synchronize()
 
         self._allreduce_futures = []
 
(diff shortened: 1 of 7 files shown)


torchft/manager.py
```diff
@@ -324,9 +349,8 @@ def allreduce(
             # Run the allreduce async and save the work object so we can wait on
             # it later.
             fut: Optional[
-                torch.futures.Future[None]
+                torch.futures.Future[List[torch.Tensor]]
                 | torch.futures.Future[torch.Tensor]
-                | torch.futures.Future[List[torch.Tensor]]
             ] = None
             if should_quantize and IS_TRITON_AVAILABLE:
                 fut = allreduce_quantized([tensor], ReduceOp.AVG, self._pg)
```
d4l3k: slightly unrelated to your PR but AVG is using world size right and not num_participants? Wonder if that's an issue

torchft/manager.py
```diff
@@ -686,21 +707,24 @@ def should_commit(self, timeout: Optional[timedelta] = None) -> bool:
         Raises:
             RuntimeError: if should_commit fails max_retries times in a row and max_retries is set
         """
-        for work in self._pending_work:
-            # check at the beginning of since .wait() may trigger errors
-            if self._errored is not None:
+        while True:
+            if len(self._pending_work) == 0:
```
d4l3k:
```
while self._pending_work:
```

torchft/local_sgd.py
```diff
@@ -324,18 +322,27 @@ def _average_grads(self) -> None:
...
             else:
-                work = self._manager.allreduce(
-                    p.grad, should_quantize=self.should_quantize
-                )
-            self._allreduce_futures.append(work)
+                tensors.append(p.grad)
+
+        work = self._manager.collect_all_allreduce(
+            tensors, should_quantize=self.should_quantize
+        )
```
d4l3k: what's the benefit for this? just to simplify the code some?

torchft/manager.py
```diff
@@ -288,8 +288,33 @@ def shutdown(self, wait: bool = True) -> None:
             self._manager.shutdown()
         self._executor.shutdown(wait=wait)
 
+    def collect_all_allreduce(
+        self, tensors: List[torch.Tensor], should_quantize: bool = False
+    ) -> torch.futures.Future[List[torch.futures.Future[torch.Tensor]]]:
```
d4l3k: If we're doing a bulk operation we should consider using `allreduce_coalesced` instead of N individual collective operations and matching the naming scheme

torchft/local_sgd.py
tushar00jain: nah this is needed because of the change in manager. it only waits for 1 piece of work before allowing to commit. 

i'm planning to have `allreduce` implement the same logic as `collect_all_allreduce` and remove `collect_all_allreduce`

torchft/manager.py
tushar00jain: yeah changed to using `SUM` since that's supported now

torchft/manager.py
tushar00jain: as discussed `_pending_work` is not needed in manager anymore

torchft/local_sgd.py
tushar00jain: this is also not needed now since `_pending_work` is not needed in manager anymore

train_diloco.py
```diff
@@ -0,0 +1,226 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
H-Huang, line 1: update `README.md` with instructions to run this script as well?

torchft/manager.py
```diff
@@ -331,34 +331,36 @@ def allreduce(
...
             # Run the allreduce async and save the work object so we can wait on
             # it later.
-            fut: Optional[
-                torch.futures.Future[None]
-                | torch.futures.Future[torch.Tensor]
-                | torch.futures.Future[List[torch.Tensor]]
-            ] = None
             if should_quantize and IS_TRITON_AVAILABLE:
-                fut = allreduce_quantized([tensor], ReduceOp.AVG, self._pg)
+                assert False, "allreduce_quantized is not supported yet"
```
H-Huang, line 335: how come the quantized version isn't supported? I think you can pass in your own stream to `allreduce_quantized` as well

torchft/local_sgd.py
```diff
@@ -606,16 +615,20 @@ def _step_post_hook(
             step = self._local_step
 
         # Start sending fragments
-        for fragment in self._fragments:
+        for i, fragment in enumerate(self._fragments):
             if not fragment.should_prepare_fragment(step):
                 continue
 
+            logger.info(f"preparing fragment {i} at step {step}")
```
H-Huang: i imagine these logs might get noisy, maybe set to debug

H-Huang (approved):
I think we can also move Diloco into its own file rather than having it in `local_sgd.py`, can be done in a later PR

torchft/manager.py
tushar00jain: `allreduce_quantized` calls `wait` on the `Work` object internal to `ProcessGroupNCCL`. This `wait` calls `synchronize` on a stream internal to `ProcessGroupNCCL` so will block the cpu until all work on that internal stream is done.

PR #210, opened 2025-06-09
Reduce exploration space in quantization and collectives tests

dzmitry-huba:
In order to speed up CI this change reduces exploration space of the quantization and collectives tests while keeping the coverage high.


dzmitry-huba:
I tried to spread the combination of different tensor sizes across quantization and collectives tests (collectives call into quantization) so the overall reduction should not be significant (though no easy way to estimate)

PR #211, opened 2025-06-10
support async in nccl pg

tushar00jain:
Summary:
- set the same stream as the one used for work in future continuations so that random streams don't depend on pg stream (this can make these streams dependent on the allreduce stream)
- wait on the work sent to pg's immediately on the fragment streams (used for allreduce) to make them depend on the pg stream and so that they don't depend on any future work that's submitted to those streams
- copy grads before allreduce so that the inner optimization can use those and it doesn't create a dependency between the default stream and the pg stream
- add back support for quantized allreduce in manager
- change return types to be consistent with pg allreduce
- the returned future from quantization collectives hangs (likely because set_result is not called?) so changed it to return the future directly from the pg

Test Plan:
- tested the changes with nccl pg
- synchronize on recovery stream sometimes makes the cpu block on collective (probably because some callback gets scheduled on the recovery stream? we need to remove synchronizing on recovery stream when there is no need to)
- calling `work.wait` returned by baby nccl pg makes the cpu block on the collective (because 2 contexts can't overlap?)
- pg gloo needs us to call `future.wait` in the sync phase instead of the prepare phase, so we probably need a different wrapper
- same for baby gloo pg

> Without Quantization

image

> With Quantization

image

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/211).
* #212
* #213
* __->__ #211

### DIFF
--- a/torchft/collectives.py
+++ b/torchft/collectives.py
@@ -135,21 +135,24 @@ def allocate_reduce_scatter_output(
     return tensor, padded_sizes
 
 
-class _QuantizedOpFuture(Future[None]):
+class _QuantizedOpFuture(Future[list[torch.Tensor]]):
     def __init__(
         self,
         sync_stream: cuda.Stream,
         keep_alive_tensors: list[torch.Tensor],
+        return_tensors: list[torch.Tensor],
     ) -> None:
         super().__init__()
         self._sync_stream = sync_stream
         self._keep_alive_tensors = keep_alive_tensors
+        self._return_tensors = return_tensors
 
-    def wait(self) -> None:
+    def wait(self) -> list[torch.Tensor]:
         # Wait for the synchronization to complete.
         cuda.current_stream().wait_stream(self._sync_stream)
         # Clean up intermediate buffers.
         del self._keep_alive_tensors
+        return self._return_tensors
 
 
 def reduce_scatter_quantized(
@@ -276,6 +279,7 @@ def reduce_scatter_quantized(
                 quantized_inputs,
                 quantized_inputs_out,
             ],
+            [output],
         )
 
 
@@ -284,7 +288,7 @@ def allreduce_quantized(
     opts: AllreduceOptions | ReduceOp,
     process_group: "ProcessGroup",
     sync_stream: cuda.Stream | None = None,
-) -> Future[None]:
+) -> Future[list[torch.Tensor]]:
     """
     Performs a quantized all-reduce operation on a list of tensors.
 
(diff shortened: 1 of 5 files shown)


torchft/collectives.py
```diff
@@ -370,20 +373,22 @@ def allreduce_quantized(
...
         # Collect reduced chunks from other ranks.
-        process_group.allgather_into_tensor_coalesced(
+        work = process_group.allgather_into_tensor_coalesced(
             [quantized_tensors.view(world_size, -1)],
             [torch.split(quantized_tensors_out.view(world_size, -1), 1)[rank]],
             _to_allgather_options(allreduce_opts),
-        ).wait()
+        )
+        work.wait()
+        fut = work.get_future()
```
d4l3k, line 383: if we call work.wait() there's no need for get_future, just run the callback inline

torchft/manager.py
```diff
@@ -409,26 +409,31 @@ def wrap_future(
         Args:
             fut: the Future to wrap
             default: the default value to complete the Future with if an error occurs
+            stream: the stream to run the continuation on, if None, the default stream will be used
```
d4l3k: nit: remove since this arg doesn't exist

torchft/collectives.py
```diff
@@ -135,21 +135,23 @@ def allocate_reduce_scatter_output(
...
         self._sync_stream = sync_stream
         self._keep_alive_tensors = keep_alive_tensors
+        self._return_tensors = return_tensors
 
-    def wait(self) -> None:
+    def wait(self) -> list[torch.Tensor]:
         # Wait for the synchronization to complete.
         cuda.current_stream().wait_stream(self._sync_stream)
         # Clean up intermediate buffers.
-        del self._keep_alive_tensors
```
H-Huang, line 152: how come this del was removed?

torchft/collectives.py
tushar00jain: we're getting the fut here so that we can return it

tushar00jain:
@H-Huang there's also a 4th stream involved that runs the callbacks

1. computation stream
2. (torch.distributed) collective stream
3. diloco fragment stream (many)
4. callback stream

i think this is how it works

## Before

- 3 waits for 2
- 4 waits for 2
- 3 waits for 4
- we sync 3' (for another fragment)
- we sync 1

3 will block cpu if `4==3'` or `4==1` 

## After

we enforce `4==3`, so

- 3 waits for 2
- we sync 3'
- we sync 1

PR #212, opened 2025-06-10
enable merging parameters for diloco

tushar00jain:
Summary:
- merge local and global parameters of the model after synchronization
- add the "alpha" parameter to integration tests

Test Plan:
```
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/212).
* #230
* __->__ #212

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -213,7 +213,13 @@ class _StreamingDiLoCoFragment:
         self.should_quantize = should_quantize
 
         self._grads: Dict[str, torch.Tensor] = {}
+
+        # Used to save global parameters so that they can be restored in case
+        # commit fails
         self.original_parameters: Dict[str, torch.Tensor] = {}
+
+        # Used to mix the local and global parameters
+        self._local_parameters: Dict[str, torch.Tensor] = {}
 
         for name, p in self._model_fragment.named_parameters():
             if isinstance(p, DTensor):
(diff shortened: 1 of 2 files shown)


torchft/local_sgd_integ_test.py
```diff
@@ -719,6 +721,7 @@ def test_streaming_diloco_commit_failure(
                     "diloco_args": {
                         "fragment_sync_delay": fragment_sync_delay,
                         "sync_every": 4,
+                        "fragment_update_alpha": alpha,
```
d4l3k, line 724: Thoughts on adding some mocked out tests where we actually check the numerical values against some references to avoid regressions?

With a fixed seed and deterministic mode in `torch` we can add regression tests by comparing against a known value.

torchft/local_sgd.py
```diff
@@ -293,6 +307,21 @@ def _set_grads(self) -> None:
...
+        Clears the saved copy of the model's parameters
+        """
+        self._local_parameters = {}
+
+    def _merge_parameters(self) -> None:
+        """
+        Merges the local and global parameters.
+        """
+        for name, p in self._model_fragment.named_parameters():
+            torch.lerp(
```
d4l3k: this isn't inplace right? having some numerics tests would be nice to catch issues like this

tushar00jain: yep let me add this and regression test in a separate diff?

PR #215, opened 2025-06-15
fix gradient allreduce

tushar00jain:
Summary:
- fix setting `_local_tensor` of a dtensor directly
- fix allreduce bucketized to not use `parameter.grad`
- simplify some code

Test Plan:
- added a test to validate the gradient are saved and set correctly
- the previous test in `local_sgd_test` fails because allreduce is not performed on `param.grad`
- updated the test to first set the grads, then load the grads to make sure they reflect the allreduce result

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/215).
* #212
* #217
* #216
* __->__ #215

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -257,17 +257,41 @@ class _StreamingDiLoCoFragment:
                 else:
                     p.data.copy_(self.original_parameters[name], non_blocking=False)
 
+    def _save_grads(self) -> None:
+        """
+        Saves pseudo-gradients of the parameters
+        """
+        with torch.no_grad():
+            for name, p in self._model_fragment.named_parameters():
+                if isinstance(p, DTensor):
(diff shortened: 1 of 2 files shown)


torchft/local_sgd.py
```diff
@@ -257,17 +257,38 @@ def restore_parameters(self) -> None:
...
         """
         Sets the gradients of the model fragment from the allreduce result
         """
-        for name, p in self._model_fragment.named_parameters():
-            if isinstance(p, DTensor):
-                p.grad._local_tensor = self._grads[name]
-            else:
-                p.grad = self._grads[name]
-
-            del self._grads[name]
```
H-Huang, line 270: how come `del` was removed?

H-Huang (approved):

PR #216, opened 2025-06-15
avoid stream synchronization in manager

tushar00jain:
Summary:
- use a recovery event to synchronize on instead of the recovery stream
- fix calling `work.wait()` in manager
- avoid calling `quorum.wait` inside of a callback

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -219,6 +219,9 @@ class Manager:
             torch.cuda.Stream() if torch.cuda.is_available() else None
         )
 
+        # Used to synchronize recovery operation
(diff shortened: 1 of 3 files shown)


torchft/manager.py
```diff
@@ -323,6 +326,7 @@ def allreduce(
             return fut
 
         self.wait_quorum()
```
H-Huang, line 328: do we need this `wait_quorum()` if `self.num_participants()` already calls it?

tushar00jain: i think good to leave it here for readability

PR #217, opened 2025-06-16
fix infinite recovery

tushar00jain:
Summary:
- we don't increase the max_step when a node is catching up because we don't call should_commit
- this can lead the node always being behind and get into an infinite recovery loop
- so simply call `should_commit`
- note, this can result in the global parameters falling out of sync, the diff includes an RFC on how to fix that

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -257,17 +257,35 @@ class _StreamingDiLoCoFragment:
                 else:
                     p.data.copy_(self.original_parameters[name], non_blocking=False)
 
+    def _save_grads(self) -> None:
+        with torch.no_grad():
+            for name, p in self._model_fragment.named_parameters():
+                if isinstance(p, DTensor):
+                    local_param = p.to_local()
+                else:
+                    local_param = p
+                pseudogradient = local_param - self.original_parameters[name].to(
+                    p.device
+                )
+                self._grads[name] = pseudogradient
+
     def _set_grads(self) -> None:
         """
         Sets the gradients of the model fragment from the allreduce result
         """
-        for name, p in self._model_fragment.named_parameters():
-            if isinstance(p, DTensor):
-                p.grad._local_tensor = self._grads[name]
-            else:
-                p.grad = self._grads[name]
-
-            del self._grads[name]
+        with torch.no_grad():
+            for name, p in self._model_fragment.named_parameters():
+                # avoid copying the gradient, it should be on the same device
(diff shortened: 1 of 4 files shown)


d4l3k (approved):
LGTM, would be nice to add some mocked test for this case if possible to prevent regressions

torchft/local_sgd.py
```diff
@@ -708,6 +734,16 @@ def _step_post_hook(
             # waste after recovery
             self._quorum_loop()
 
+            # TODO: Since we do quorum after commit, there might be a big gap until
+            # the next allreduce. This increases the chances of nodes failing
+            # and so the allreduce to fail.
+            # - We could maybe do a quorum again right before preparing for a fragment
+            #   using `shring_only`. This might make it tricky for new nodes to join
```
H-Huang: typo

torchft/local_sgd.py
```diff
@@ -559,7 +592,7 @@ def __init__(
             _StreamingDiLoCoFragment(
                 manager,
                 model_fragment,
-                math.floor((sync_every / len(model_fragments)) * (i + 1)),
+                (sync_every // len(model_fragments) * (i + 1)),
```
d4l3k: Should we assert that this is an exact multiple of model_fragments? I suppose it doesn't matter too much if it's not exact but might surprise people?

d4l3k (approved):

torchft/local_sgd.py
```diff
@@ -660,12 +702,16 @@ def _step_post_hook(
...
             logger.debug(f"preparing fragment {i} at step {step}")
 
+            self._first_prepare_sent.add(i)
             fragment.prepare_sync()
 
         for i, fragment in enumerate(self._fragments):
             if not fragment.should_sync_fragment(step):
                 continue
 
+            if i not in self._first_prepare_sent:
```
H-Huang, line 714: can you add a comment when we would get into this case? It looks like we set it in the loop before so i'm confused.

torchft/local_sgd.py
tushar00jain: i feel more asserts will require people to remember more things on how to tune the settings? don't have a strong pref if you think that's easier to reason about though, happy to make changes later as well

torchft/local_sgd.py
tushar00jain: the loop before could run for a different fragment (not the one we're syncing) depending on the sync schedule

PR #220, opened 2025-06-20
document protocol and recommended setup

tushar00jain:
Summary:
added 2 pages,

- "protocol" page describes the protocol
- "assumptions-and-recommendations" describes assumptions under which garuntees given by the protocol hold and some configuraitons that users should know about when running torchft

Test Plan:
```
make livehtml
```

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -11,7 +11,14 @@ the entire training job.
 
 .. toctree::
     :maxdepth: 1
-    :caption: Reference
+    :caption: Design
+
+    protocol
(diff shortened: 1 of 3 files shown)


docs/source/index.rst
```diff
@@ -13,6 +13,13 @@ the entire training job.
     :maxdepth: 1
     :caption: Reference
```
d4l3k: Maybe rename this as Usage or Design?

d4l3k (approved):
LGTM

PR #221, opened 2025-06-23
docs: pytorch-sphinx-theme2

d4l3k:
This updates the torchft docs to use the new pytorch-sphinx-theme2.

https://github.com/pytorch/pytorch_sphinx_theme/blob/pytorch_sphinx_theme2/pytorch_sphinx_theme2_migration_guide.md

Test plan:

```
make html
```


docs/source/index.rst
```diff
@@ -17,18 +17,11 @@ the entire training job.
     assumptions_and_recommendations
 
 .. toctree::
-    :maxdepth: 1
+    :maxdepth: 2
```
svekars, line 20: so what happens if set this to 1?
```suggestion
    :maxdepth: 1
```

d4l3k: This doesn't help it just only shows "API Reference"

api.md is already set to 1

svekars: can you set this maxdepth to 1? and remove the caption.

tushar00jain: yeah will be nice to have it like we did before, index page looks a bit clunky now

docs/requirements.txt
```diff
@@ -1,9 +1,15 @@
...
+sphinxcontrib.katex==0.8.6
+sphinxext-opengraph==0.9.1
+#breathe==4.34.0  # only if generating C++
+#exhale==0.2.3  # only if generating C++ docs
+docutils==0.16
+sphinx-design==0.4.0
+sphinxcontrib-mermaid==1.0.0
+#myst-parser==0.18.1  # if want to contribute in markdown
+#sphinx-gallery==0.11.1  # if hosting interactive tutorials
+
```
svekars, line 14:
```suggestion
sphinx-sitemap==2.7.1
```

docs/source/index.rst
d4l3k: @svekars We want the sub pages to show so we do want to keep this at maxdepth 2. I don't want to make users have to click through two pages

docs/requirements.txt
d4l3k: Ah, this isn't in the migration guide

d4l3k:
Docs build is green -- landing

PR #222, opened 2025-06-23
add streaming diloco test - recovery

tushar00jain:
Summary:
- update diloco tests to work with model fragments
- propogate arguments for training loop and optimizer wrapper through test runner
- added a test that validates streaming diloco works when node crashes and rejoins
- store the whole manager state dict for validation instead of only user's default state dict
- fixed a bug that made the local step go out of sync

Test Plan:
```
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/222).
* #223
* __->__ #222

### DIFF
--- a/torchft/local_sgd_integ_test.py
+++ b/torchft/local_sgd_integ_test.py
@@ -2,13 +2,13 @@ import copy
 import logging
 import os
 import re
+import sys
 import traceback
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import ExitStack
 from datetime import timedelta
-from sys import platform
 from typing import Any, Dict
-from unittest import TestCase
+from unittest import TestCase, skipIf
 
 import torch
 from parameterized import parameterized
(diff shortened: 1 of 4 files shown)


torchft/local_sgd_integ_test.py
```diff
@@ -382,7 +426,7 @@ def test_diloco_recovery(self, use_cuda: bool) -> None:
 
         torch.manual_seed(42)
         # Initialize the model so we can pass in the state_dict
-        m: nn.Module = MyModel(2, 3)
+        m: nn.Module = MultiMyModel(2, 3, 1)
```
d4l3k, line 429: is it worth testing the non-streaming case as well?

d4l3k (approved):
LGTM

torchft/local_sgd_integ_test.py
tushar00jain: this test case is non-streaming? next one is for streaming

PR #223, opened 2025-06-23
add streaming diloco tests - upscale

tushar00jain:
Summary:
- add a "barrier injector" to allow replicas to join at specific steps
- added a test to validate streaming diloco works when new node joins without any failures
-  remove some code duplication

Test Plan:
```
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/223).
* #224
* __->__ #223

### DIFF
--- a/torchft/local_sgd_integ_test.py
+++ b/torchft/local_sgd_integ_test.py
@@ -3,9 +3,11 @@ import logging
 import os
 import re
 import sys
+import threading
 import traceback
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import ExitStack
+from dataclasses import field
 from datetime import timedelta
 from typing import Any, Dict
 from unittest import TestCase, skipIf
@@ -13,13 +15,14 @@ from unittest import TestCase, skipIf
 import torch
 from parameterized import parameterized
 from torch import nn, optim
+from torch.distributed.pipelining import SplitPoint, pipeline
 from torch.distributed.tensor import DTensor, Replicate
 
 from torchft._torchft import LighthouseServer
 from torchft.device_mesh import ft_init_device_mesh
 from torchft.local_sgd import DiLoCo, LocalSGD
 from torchft.manager import Manager
-from torchft.manager_integ_test import FailureInjector, MyModel, Runner
+from torchft.manager_integ_test import BarrierInjector, FailureInjector, MyModel, Runner
 from torchft.process_group import ProcessGroupBabyNCCL, ProcessGroupGloo
 
 logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


torchft/local_sgd_integ_test.py
```diff
@@ -562,14 +584,93 @@ def test_streaming_diloco_recovery(self, use_cuda: bool) -> None:
...
+    CONFIG: list[tuple[bool, int, int]] = [
+        (use_cuda, n_fragments, fragment_sync_delay)
+        for use_cuda in [True, False]
+        for n_fragments in [1, 2]
+        for fragment_sync_delay in [0, 1]
+    ]
+
+    # pyre-fixme[56]: Pyre was not able to infer the type of argument
+    @skipIf(sys.platform == "darwin", "not reliable on mac")
+    @parameterized.expand(CONFIG)
```
d4l3k, line 598: usually we just put these inline

torchft/manager_integ_test.py
```diff
@@ -83,6 +83,30 @@ def check(self, rank: int, step: int) -> None:
                 raise InjectedFailure(f"injected failure {rank=} {step=}")
 
 
+class BarrierInjector:
```
d4l3k, line 86: thoughts on making a generic Injector class with several hooks that can be overridden?

torchft/local_sgd_integ_test.py
```diff
@@ -562,14 +584,93 @@ def test_streaming_diloco_recovery(self, use_cuda: bool) -> None:
...
+            min_replicas=2,
+        )
+        num_replicas = 3
+        futures = []
+        executors = []
+
+        barrier = threading.Barrier(num_replicas)
+
+        barrier_injectors = [
+            # Make this replica join after other replicas have made 2 steps
```
d4l3k, line 617: took me a second to understand how this works

Does this run into any issues given the join window timeout since the worker will be immediately heartbeating even though it doesn't join until step 2?

torchft/local_sgd_integ_test.py
tushar00jain: linter complains about not being able to figure out the type

torchft/local_sgd_integ_test.py
tushar00jain: yeah the quorum timeout needs to be set high enough that this node doesn't crash

torchft/manager_integ_test.py
tushar00jain: yeah that'd be good, maybe separate diff though?

torchft/local_sgd_integ_test.py
tushar00jain: also join_timeout is 100ms (https://github.com/tushar00jain/torchft/blob/e5dfb438f2a823b0556ed13cbb49496a62900c93/src/lib.rs#L623

torchft/manager_integ_test.py
tushar00jain: https://github.com/pytorch/torchft/pull/224

PR #226, opened 2025-06-24
test allreduce failures for diloco

tushar00jain:
Summary:
- test when allreduce fails but no new nodes join

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -268,6 +268,15 @@ class Manager:
         self._participating_replica_rank: Optional[int] = None
         self._participating_replica_world_size: int = 0
 
+        # used to artificially fail the next allreduce by tests
+        self._TEST_should_fail_allreduce = False
+
+    def TEST_fail_allreduce(self) -> None:
+        """
+        Fails the next allreduce. This is used for testing.
+        """
+        self._TEST_should_fail_allreduce = True
+
     def register_state_dict_fn(
         self,
         key: str,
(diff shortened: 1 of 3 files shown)


torchft/manager.py
```diff
@@ -268,6 +268,15 @@ def __init__(
         self._participating_replica_rank: Optional[int] = None
         self._participating_replica_world_size: int = 0
 
+        # used to artificially fail the next allreduce by tests
+        self._TEST_should_fail_allreduce = False
+
+    def TEST_fail_allreduce(self) -> None:
```
d4l3k: can we create a wrapped/mocked PG that injects the failure instead?

tushar00jain: yeah let me see if we can create a wrapper so we can keep using real pg. thinking of using mocked pg for deterministic simulation

H-Huang: agree that we should try to keep test related functionality contained in our test files!

torchft/process_group.py
```diff
@@ -1016,6 +1023,55 @@ def allreduce(self, tensors: List[torch.Tensor], opts: object) -> Work:
             return _DummyWork(tensors)
 
 
+class FakeProcessGroupWrapper(ProcessGroupWrapper):
```
H-Huang, line 1026: nit: could maybe move this class into tests?

H-Huang (approved):
LGTM

torchft/process_group.py
tushar00jain: yeah there's a bunch of other fakes in this file. maybe take this up separately? we also want to upstream process_group to pytorch

PR #228, opened 2025-07-10
retry quorum

tushar00jain:


src/manager.rs
```diff
@@ -226,6 +222,45 @@ impl Manager {
 
         Ok(())
     }
+
+    async fn _quorum_with_retries(
+        &self,
+        timeout: Duration,
+        lighthouse_request: LighthouseQuorumRequest,
+    ) -> Result<tonic::Response<LighthouseQuorumResponse>, Status> {
+        let mut client = self.lighthouse_client.clone();
```
d4l3k: Will this reconnect if an error occured or do we need to construct a new lighthouse_client in that case?

src/manager.rs
```diff
@@ -226,6 +222,45 @@ impl Manager {
...
+                Err(e) => {
+                    info_with_replica!(
+                        self.replica_id,
+                        "lighthouse quorum timed out: {}",
+                        e.to_string()
+                    );
+
+                    if retry_count == self.quorum_retries {
+                        return Err(Status::internal(format!(
+                            "lighthouse quorum timed out after {} retries",
```
d4l3k: Should we include the last error as well?

src/manager.rs
```diff
@@ -226,6 +222,45 @@ impl Manager {
...
+            let result = tokio::time::timeout(timeout, client.quorum(request)).await;
+
+            match result {
+                Ok(response) => {
+                    return response;
+                }
+                Err(e) => {
+                    info_with_replica!(
+                        self.replica_id,
+                        "lighthouse quorum timed out: {}",
```
d4l3k: This could be non-timeouts as well right? Connection resets etc would also hit this path

torchft/manager.py
```diff
@@ -271,6 +276,7 @@ def __init__(
                 world_size=group_world_size,
                 heartbeat_interval=heartbeat_interval,
                 connect_timeout=connect_timeout,
+                quorum_retries=int(os.environ.get(QUORUM_RETRIES_ENV, "0")),
```
d4l3k: may be good to also add a config variable for this

d4l3k (approved):
LGTM but have some comments. If we want to target things other than just timeouts I think we need to tweak the retry loop

torchft/manager.py
tushar00jain: the env var is the config? or you mean somewhere else?

src/manager.rs
tushar00jain: updated, now we create a new client each time the request fails

torchft/manager.py
d4l3k: i.e. able to specify it via an arg to Manager init

src/manager.rs
```diff
@@ -170,52 +180,50 @@ impl Manager {
...
             });
 
-            let _response = client.heartbeat(request).await;
+            if let Err(e) = client.heartbeat(request).await {
+                info_with_replica!(
+                    self.replica_id,
+                    "Failed to send heartbeat to lighthouse: {}",
+                    e.to_string()
+                );
+                let _ = self.create_lighthouse_client().await;
```
d4l3k, line 199: Should we be checking status for this?

d4l3k (approved):
LGTM

src/manager.rs
tushar00jain: it's fine I think, if we were unable to replace the lighthouse client, next attempt to send the heartbeat will also fail and it'll again try to create a lighthouse client

torchft/manager.py
tushar00jain: done

PR #229, opened 2025-07-10
make timeouts configurable

tushar00jain:
Summary:
while training, we need to set higher quorum timeouts to make sure all replicas can finish training

these parameters are only exposed through the manager but users may not be able to access the manager directly e.g. when using torchtitan

so make the timeouts configurable using env vars that take precedence over the parameters passed through manager

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/229).
* #228
* __->__ #229

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -61,7 +61,31 @@ MANAGER_ADDR_KEY: str = "manager_addr"
 MANAGER_PORT_ENV: str = "TORCHFT_MANAGER_PORT"
 REPLICA_ID_KEY: str = "replica_id"
 
+# Environment variables for various timeouts. These can also be passed
+# in through the manager but the environment variables take precedence.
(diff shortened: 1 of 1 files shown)


d4l3k (approved):
LGTM

torchft/manager.py
```diff
@@ -177,9 +183,29 @@ def __init__(
...
+        else:
+            self._quorum_timeout: timedelta = quorum_timeout
+
+        connect_timeout_sec_env = os.environ.get(CONNECT_TIMEOUT_SEC_ENV, None)
+        if connect_timeout_sec_env is not None:
+            self._connect_timeout: timedelta = timedelta(
+                seconds=int(connect_timeout_sec_env)
+            )
+        else:
+            self._connect_timeout: timedelta = connect_timeout
```
d4l3k: could make this into a helper function to reduce duplication

PR #232, opened 2025-07-14
refactor diloco test

tushar00jain:
Summary:
- move the training loop to a separate file
- convert it into a class so that methods can be overridden without having to duplicate code

### DIFF
--- a/torchft/diloco_trainer.py
+++ b/torchft/diloco_trainer.py
@@ -1,0 +1,313 @@
+import copy
+import logging
+import os
+from contextlib import ExitStack
+from datetime import timedelta
+from typing import Any, Dict, List, cast
+
+import torch
+from numpy import outer
+from torch import nn, optim
+from torch.distributed.tensor import DTensor
+from torch.optim.optimizer import Optimizer
+from torch.optim.sparse_adam import F
+
+from torchft.device_mesh import ManagedDeviceMesh, ft_init_device_mesh
+from torchft.local_sgd import DiLoCo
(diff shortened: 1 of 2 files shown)


torchft/diloco_trainer.py
```diff
@@ -0,0 +1,310 @@
...
+
+                self.inner_optimizer.zero_grad()
+                loss.backward()
+                self.inner_optimizer.step()
+
+                # after 4 model updates then break
+                if self.manager.current_step() >= 4:
+                    break
+
+        return all_state_dicts
```
d4l3k: Not directly related to this PR but wonder if we should just add a __main__ check and use this for manual testing

torchft/diloco_trainer.py
```diff
@@ -0,0 +1,310 @@
+import copy
```
d4l3k: Does this makes sense under torchft/? Should we add an underscore to indicate it's internal or move it under some `testing` dir or something?

Just don't want folks getting confused and importing this directly for whatever reason

H-Huang: agree with this

torchft/diloco_trainer.py
tushar00jain: what do you mean by `main` check? :)

torchft/diloco_trainer.py
tushar00jain: moved it to `test` folder. will start putting everything test related there

PR #233, opened 2025-07-14
simplify streaming diloco

tushar00jain:
Summary:
- since we made a simplifying assumption that we will only ever have 1 inflight fragment, we can simplify some of the logic particularly getting rid of the local step in manager state
- we'll just use the manager's step to determine which fragment to sync
- this also allows us to easily support heterogenous hardware by tuning the sync_every setting that will make slower/faster machines to perform less/more local steps before they sync
- we can also perform quorum right before preparing a fragment sync - this easily ensures that all replicas will have the same max step and sync the same fragment
- fix some numeric issues
  - the sign of the pseudogradient
  - inplace lerp when mixing local and global parameters

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/233).
* #234
* __->__ #233
* #232

### DIFF
--- a/torchft/local_sgd.py
+++ b/torchft/local_sgd.py
@@ -281,8 +281,8 @@ class _StreamingDiLoCoFragment:
                     local_param = p.to_local()
                 else:
                     local_param = p
-                pseudogradient = local_param - self.original_parameters[name].to(
-                    p.device
+                pseudogradient = (
+                    self.original_parameters[name].to(p.device) - local_param
                 )
                 self._grads[name] = pseudogradient
 
@@ -318,7 +318,7 @@ class _StreamingDiLoCoFragment:
         Merges the local and global parameters.
         """
         for name, p in self._model_fragment.named_parameters():
-            p.data.lerp(self._local_parameters[name], 1 - self._fragment_update_alpha)
+            p.data.lerp_(self._local_parameters[name], self._fragment_update_alpha)
 
     @torch.profiler.record_function("torchft::local_sgd::wait")
     def wait(self) -> None:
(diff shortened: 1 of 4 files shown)


torchft/local_sgd.py
```diff
@@ -281,8 +281,8 @@ def _save_grads(self) -> None:
                     local_param = p.to_local()
                 else:
                     local_param = p
-                pseudogradient = local_param - self.original_parameters[name].to(
-                    p.device
+                pseudogradient = (
+                    self.original_parameters[name].to(p.device) - local_param
```
d4l3k, line 285: This is flipped because we don't do 1- below?

torchft/local_sgd.py
```diff
@@ -588,7 +553,11 @@ def __init__(
         if sync_every < len(model_fragments):
             raise ValueError("Only 1 fragment can be syncrhonized at a time")
 
-        if fragment_sync_delay >= sync_every:
+        if sync_every % len(model_fragments) != 0:
+            raise ValueError("sync_every must divide the number of fragments")
```
d4l3k, line 557: This makes sense for now -- we can relax this later if it turns out people want to sync different parts of the model at different rates though that has other significant considerations

tushar00jain: we can still sync at different rates by passing in a different `sync_every`. will need to make it configurable in torchtitan too.

torchft/local_sgd.py
tushar00jain: what do you mean by `1-`? the outer optimizer will do `param = param - pseudo_grad`. but loss goes down in the direction of `-pseudo_grad`.  this is pretty much why we were seeing the loss going up earlier i think.

tushar00jain: proof of this:

- let's say `pseudo_grad = new_param - old_param` where `new_param = old_param - grad`
- outer optimizer step is `new_param = old_param - pseudo_grad = old_param - (old_param - grad) = old_param + grad`
- this is incorrect because it should just be `new_param = old_param - grad`

PR #234, opened 2025-07-14
regression test

tushar00jain:
Summary:
- add a test that uses fixtures to validate against previous runs of the test
- the fixutres can be written using `WRITE_FIXTURE=true pytest -vs ./torchft/test_diloco_mocked_updates.py`
- when writing fixtures, the test also numerically validates the implementation of streaming diloco


torchft/fixtures/torchft.test_diloco_mocked_updates.DiLoCoMockedUpdateTest.test_diloco_mocked_updates_0.json
d4l3k, line 1: Should we move these to a /test/* dir? It might get real annoying to grep through torchft/ if there's a bunch of testing specific autogenned files

torchft/fixtures/torchft.test_diloco_mocked_updates.DiLoCoMockedUpdateTest.test_diloco_mocked_updates_3.json
```diff
@@ -0,0 +1,546 @@
+[
+  [
+    {
+      "history": {
+        "0": {
+          "layers.0.weight": [
+            [
+              1.0
```
d4l3k: is it expected that all of these are whole numbers? I would expect the weights to be a lot more varied?

tushar00jain: yeah i'm optimizing for human readability e.g. i was able to run the numerics by hand and validate the output in these files myself. this helped in writing the validation logic and it caught issues that i fixed in the previous diff. also using floats might need us to do some closeness check because of precision issues (different machines might do things differently). i think we do get full validation of the logic using just whole numbers though?

torchft/fixtures/torchft.test_diloco_mocked_updates.DiLoCoMockedUpdateTest.test_diloco_mocked_updates_0.json
tushar00jain: done

d4l3k: ah I was thinking top level not under the torchft dir

I often find stuff by just running `rg foo torchft` so was hoping to have it excluded

tushar00jain: - moved to `/test_fixtures`
- keep the test file with the implementation file
- renamed `test` folder to `_test`

PR #237, opened 2025-07-18
regression test for failures

tushar00jain:
Summary:
- add a regression test in which 1 node crashes then restarts
- write the fixutre for this test
- make the code that checks fixutres reusable

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/237).
* #238
* __->__ #237


torchft/diloco_regression_test.py
```diff
@@ -372,6 +352,149 @@ def test_diloco_mocked_updates(
...
+
+        lighthouse = LighthouseServer(bind="[::]:0", min_replicas=2)
+        sync_every = 6
+        num_replicas = 2
+        futures = []
+
+        # Create event injectors - make the second replica fail at step 2
+        event_injectors = [
+            EventInjector(),  # First replica runs normally
+            EventInjector().fail_at(0, 2),  # Second replica fails at step 2
```
d4l3k, line 385: It's a bit funny since we recover in a single step so the weights are identical on both. That is correct behavior but wondering if we should also check the weights at the start of the step to show they're different

torchft/diloco_regression_test.py
```diff
@@ -372,6 +352,149 @@ def test_diloco_mocked_updates(
...
+                runner = Runner(
+                    replica_id=replica_id,
+                    num_replicas=num_replicas,
+                    lighthouse_address=lighthouse.address(),
+                    event_injector=event_injector,
+                    train_loop=mock_diloco_train_loop,
+                    use_cuda=use_cuda,
+                    train_loop_args={
+                        "n_fragments": n_fragments,
+                        "model_state_dict": model_state_dict,
```
d4l3k, line 404: Can we double check whether the model copy the weights from this array or are they shared between the two models?

d4l3k (approved):

torchft/diloco_regression_test.py
tushar00jain: we can just clean this up, this state dict doesn't end up being used anyway

PR #238, opened 2025-07-18
handle original parameters in diloco

tushar00jain:
Summary:
- external integration e.g. torchtitan don't have access to global parameters, which is diloco's internal state so they don't handle loading it or saving it
- we need to do this though so that the gradient computations can be correct when a node recovers
- set the global parameters inside diloco itself

Test Plan:
- the regression tests pass
- ran streamig diloco on 25 nodes with 1 fragment, syncing every 10 steps, with 2 step sync delay injecting failure every 7.5 minutes

Screenshot 2025-07-18 at 4 28 57 AM


d4l3k (approved):
LGTM

PR #239, opened 2025-07-22
allow using gloo from flag

tushar00jain:
Summary:
add an env var that enables using gloo for the sample training script

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/239).
* #240
* __->__ #239

### DIFF
--- a/train_diloco.py
+++ b/train_diloco.py
@@ -14,6 +14,7 @@ os.environ["CUDA_VISIBLE_DEVICES"] = str(REPLICA_GROUP_ID % 4)
 os.environ["NCCL_HOSTID"] = str(REPLICA_GROUP_ID)
 
 USE_STREAMING = os.getenv("USE_STREAMING", "False") == "True"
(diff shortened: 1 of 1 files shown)


d4l3k (approved):
LGTM

train_diloco.py
```diff
@@ -14,6 +14,7 @@
 os.environ["NCCL_HOSTID"] = str(REPLICA_GROUP_ID)
 
 USE_STREAMING = os.getenv("USE_STREAMING", "False") == "True"
+USE_NCCL = os.getenv("USE_NCCL", "False") == "True"
```
H-Huang, line 17: so we use gloo by default?

H-Huang (approved):
LGTM!

train_diloco.py
tushar00jain: yeah since nccl doesn't work very well right now

PR #240, opened 2025-07-22
use block_current_stream work api

tushar00jain:
Summary:
use block_current_stream to avoid blocking the cpu when using gloo

Test Plan:
## Before

image

## After

image

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -373,7 +373,7 @@ class Manager:
                 )
             else:
                 work = self._pg.allreduce([tensor], ReduceOp.SUM)
-                work.wait()
+                work.block_current_stream()
                 fut = work.get_future()
 
             stream: Optional[torch.cuda.Stream] = (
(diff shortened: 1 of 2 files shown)


torchft/manager.py
```diff
@@ -411,7 +412,7 @@ def callback(
             fut = fut.then(callback)
 
             fut = self.wrap_future(fut, tensor)
-            return fut
+            return (work, fut)
```
d4l3k: this makes me a bit nervous since calling only work.wait() means that the code in the future callback may have not run -- i.e `tensor /= num_participants` may execute out of order

The advantage of future objects here is that the work runs on the completing thread, where as .wait() runs on the waiting thread. I'm wondering if we should actually wrap these futures back into Work objects so we can do the stream dependency correctly in .wait()

d4l3k (approved):
LGTM

tushar00jain:
@d4l3k i was planning to combine this diff with option 1/2

torchft/manager.py
tushar00jain: wrapped the work object so this should be addressed?

PR #243, opened 2025-07-26
option 1 - use block_current to overlap compute/communication

tushar00jain:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/243).
* __->__ #243
* #240

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -26,6 +26,7 @@ and Hybrid FSDP.
 """
 
 import concurrent.futures
+import copy
 import logging
 import os
 import socket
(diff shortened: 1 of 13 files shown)


torchft/manager.py
```diff
@@ -382,7 +382,7 @@ def allreduce(self, tensor: torch.Tensor, should_quantize: bool = False) -> Work
                 )
             else:
                 work = self._pg.allreduce([tensor], ReduceOp.SUM)
-                work.wait()
+                work.block_current_stream()
```
d4l3k: this partially solves it but it doesn't really help the case below with the tensor division

Ideally we wrap the future below in a Work object and then call .block_current_stream() on that

tushar00jain: > this partially solves it but it doesn't really help the case below with the tensor division

in which case?

- for nccl with cuda, the behavior should be the same as the existing one
- for gloo with cuda, the tensor is on the gpu (after the host to device copy but the tensor arg is also on gpu) so the callback will also return immediately? iiuc the fut.wait() in the callback that i added will also return immediately
- for gloo without cuda, based on what you said the callback will be called after the device to host copy has been completed?

tushar00jain: > Ideally we wrap the future below in a Work object and then call .block_current_stream() on that

we also need to call `work.wait()` or `work.block_current_stream()` to make sure work finishes on the current stream first before the future runs on the current stream

d4l3k (approved):
LGTM

We should document in README that we require torch nightly

PR #244, opened 2025-07-26
use http transport

tushar00jain:
Summary:
use http transport instead of pg transport -- pg transport fails to resolve address when running locally

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/244).
* #248
* #243
* #240
* #247
* #246
* #245
* __->__ #244


d4l3k (approved):
LGTM

PR #245, opened 2025-07-26
make checkpointing thread safe

tushar00jain:
Summary:
- the checkpointing wasn't thread safe for http transport so use lock the model in the pre step hook and when we want to transfer the checkpoint

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -55,6 +55,7 @@ from torch.distributed.distributed_c10d import AllreduceOptions, ReduceOp, Work
 
 from torchft._torchft import ManagerClient, ManagerServer
 from torchft.checkpointing import CheckpointTransport, HTTPTransport
+from torchft.checkpointing._rwlock import RWLock
 from torchft.futures import future_timeout
 from torchft.work import _DummyWork
 
@@ -216,6 +217,9 @@ class Manager:
         self._load_state_dict_fns: Dict[str, Callable[[object], None]] = {}
         self._user_state_dicts: Dict[str, Callable[[], object]] = {}
 
+        # Protects state dict
+        self._state_dict_lock = RWLock(timeout=timeout.total_seconds())
+
         if load_state_dict and state_dict:
             self.register_state_dict_fn("default", load_state_dict, state_dict)
 
@@ -324,6 +328,21 @@ class Manager:
         # first step is 1
         self._participating_replica_rank: Optional[int] = None
         self._participating_replica_world_size: int = 0
+        self._is_state_dict_read_allowed = True
+
+    def allow_state_dict_read(self) -> None:
+        if self._is_state_dict_read_allowed:
+            return
+
+        self._is_state_dict_read_allowed = True
+        self._state_dict_lock.w_release()
+
+    def disallow_state_dict_read(self) -> None:
+        if not self._is_state_dict_read_allowed:
+            return
+
+        self._is_state_dict_read_allowed = False
+        self._state_dict_lock.w_acquire()
 
     def register_state_dict_fn(
         self,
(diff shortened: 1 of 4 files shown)


torchft/checkpointing/http_transport.py
```diff
@@ -278,7 +279,13 @@ def _to_cpu(values: List[T], pin_memory: bool) -> List[T]:
                 else:
                     out.append(v.cpu())
             else:
-                out.append(v)
+                if isinstance(v, DTensor):
+                    clone = distribute_tensor(
+                        v.to_local().clone(), v.device_mesh, v.placements
```
d4l3k: Is `v.clone()` not sufficient? What does that do? Do we also need special logic for cuda DTensors?

torchft/checkpointing/http_transport.py
```diff
@@ -278,7 +279,13 @@ def _to_cpu(values: List[T], pin_memory: bool) -> List[T]:
                 else:
                     out.append(v.cpu())
             else:
-                out.append(v)
+                if isinstance(v, DTensor):
```
d4l3k: should we rename `_to_cpu` to `_clone_cpu`?

d4l3k (commented):
Is there some other way we can handle this? 

Two main concerns:

* cloning has significant memory impact
* send_checkpoint may race with the first inner optimizer step so cloining in _to_cpu may not even be safe

I thought with the DiLoCo implementation we had a second copy of the weights to compute the pseudo gradient. Can we not reuse those for the state_dict transfer?

torchft/checkpointing/http_transport.py
tushar00jain: > Is v.clone() not sufficient?

tried it and didn't work. it made an empty dtesnor. took a long time to figure how to clone dtensors

tushar00jain:
@d4l3k 

> I thought with the DiLoCo implementation we had a second copy of the weights to compute the pseudo gradient. Can we not reuse those for the state_dict transfer?

the model transfer isn't even controlled by the wrapper in general 

> Is there some other way we can handle this?

thought about blocking everything until checkpoint transfer is complete but that's probably also complicated since the node may never want to transfer checkpoint

tushar00jain:
@d4l3k 

> I thought with the DiLoCo implementation we had a second copy of the weights to compute the pseudo gradient. Can we not reuse those for the state_dict transfer?

the model transfer isn't even controlled by the wrapper in general 

> Is there some other way we can handle this?

thought about blocking everything until checkpoint transfer is complete but that's probably also complicated since the node may never want to transfer checkpoint

tushar00jain:
@d4l3k for the gpu case, we're clone the tensor into cpu memory anyway. thing is we don't control at what inner step the checkpoint will be sent so the regression test ends up being non deterministic. it changes what model parameters are used for syncing, even if we make it thread safe

also on thread safety, the cloning seems to be a blocking call in post step hook, so it shouldn't race with the inner optimizer step?

torchft/manager.py
```diff
@@ -820,11 +839,14 @@ def load_state_dict(self, state_dict: Dict[str, int]) -> None:
         self._batches_committed = state_dict["batches_committed"]
 
     def _manager_state_dict(self) -> Dict[str, object]:
-        assert len(self._user_state_dicts) > 0, "user state_dict is not initialized."
-        return {
-            "user": {key: value() for key, value in self._user_state_dicts.items()},
-            "torchft": self.state_dict(),
-        }
+        with self._state_dict_lock.r_lock():
```
d4l3k, line 828: we already have `allow_checkpoint` and `disallow_checkpoint` in HTTPTransport -- can we reuse those instead?

tushar00jain: - that also requires copying over state dict to the http transport
- or keeping some tracking on which step was transferred last to the http transport

with a separate lock, we can decouple checkpoint specific logic with training logic

torchft/manager.py
```diff
@@ -324,6 +328,21 @@ def __init__(
         # first step is 1
         self._participating_replica_rank: Optional[int] = None
         self._participating_replica_world_size: int = 0
+        self._is_state_dict_read_allowed = True
+
+    def allow_state_dict_read(self) -> None:
```
d4l3k, line 333: Can we add specific tests for these methods? Lock logic is pretty risk prone so would be nice to have unit test coverage for these

d4l3k (approved):
Seems reasonable to me -- just want to add a unit test specifically on the lock

PR #247, opened 2025-07-26
return work from manager allreduce

tushar00jain:
Summary:
returns the work object so we can be more flexible with the usage

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/247).
* #249
* #240
* #248
* #245
* #243
* __->__ #247
* #246
* #244

### DIFF
--- a/torchft/collectives.py
+++ b/torchft/collectives.py
@@ -18,6 +18,7 @@ from torch.distributed.distributed_c10d import (
     AllreduceOptions,
     AllToAllOptions,
     ReduceScatterOptions,
+    Work,
 )
 from torch.futures import Future
 
@@ -288,7 +289,7 @@ def allreduce_quantized(
     opts: AllreduceOptions | ReduceOp,
     process_group: "ProcessGroup",
     sync_stream: cuda.Stream | None = None,
-) -> Future[list[torch.Tensor]]:
+) -> Work:
     """
     Performs a quantized all-reduce operation on a list of tensors.
 
@@ -379,10 +380,18 @@ def allreduce_quantized(
             [torch.split(quantized_tensors_out.view(world_size, -1), 1)[rank]],
             _to_allgather_options(allreduce_opts),
         )
+
+        # NOTE: This is not supposed to be used with gloo, only with NCCL.
+        # So we setup the stream dependency here by calling work.wait(),
+        # which doesn't block the CPU.
+        #
+        # The future callback below will run after the work has been
+        # completed.
+
         work.wait()
         fut = work.get_future()
 
-        def callback(fut: Future[list[torch.Tensor]]) -> list[torch.Tensor]:
+        def callback(fut: Future[list[torch.Tensor]]) -> None:
             # Dequantize and copy to output buffer.
             nonlocal tensors, quantized_tensors, world_size, sync_stream
 
@@ -391,7 +400,6 @@ def allreduce_quantized(
                 fut.wait()
                 # Dequantize the result back to the original precision
                 fused_dequantize_from_fp8(tensors, quantized_tensors, world_size)
-                return tensors
 
-        fut = fut.then(callback)
-        return fut
+        fut.add_done_callback(callback)
+        return work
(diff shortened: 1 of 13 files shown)


torchft/collectives_test.py
```diff
@@ -94,8 +94,8 @@ def _run_all_reduce_collective(
                     )
                 ]
 
-                fut = allreduce_quantized(tensors, reduce_op, pg)
-                fut.wait()
+                work = allreduce_quantized(tensors, reduce_op, pg)
+                work.get_future().wait()
```
d4l3k: why not just call .wait() here?

torchft/local_sgd.py
```diff
@@ -154,7 +155,8 @@ def _average(self) -> list[torch.Tensor]:
         for p in self._model.parameters():
             # Create a new tensor to store the averaged parameter
             avg_param = extract_local_tensor(p)
-            works.append(self._manager.allreduce(avg_param))
+            work = self._manager.allreduce(avg_param)
+            works.append(work)
```
d4l3k: any reason for this to change?

torchft/local_sgd.py
```diff
@@ -522,8 +525,10 @@ def callback(fut: torch.futures.Future[torch.Tensor]) -> None:
                             flat_buffer[pack_offset : pack_offset + numel].view_as(t)
                         )
 
-            work = work.then(callback)
-            self._allreduce_futures.append(work)
+            fut = work.get_future()
+            fut = fut.then(callback)
```
d4l3k: fut isn't consumed anywhere -- is that intentional?

torchft/work.py
```diff
@@ -0,0 +1,36 @@
...
+class _WorkWrapper(dist._Work):
+    def __init__(
+        self, work: dist._Work, fut: torch.futures.Future[torch.Tensor]
+    ) -> None:
+        super().__init__()
+        self._work = work
+        self._fut = fut
+
+    def wait(self, timeout: Optional[timedelta] = None) -> bool:
+        return True
```
d4l3k, line 34: this should block on the future

d4l3k (changes requested):

torchft/collectives_test.py
tushar00jain: work.wait should have already been called inside manager (as of this diff, it is called there). so we're waiting on the final future

tushar00jain: based on the suggestion you made, we can do work.wait(), and work.wait() calls future.wait()

torchft/local_sgd.py
tushar00jain: i was assuming this modifies the original future, that we then wait on later

torchft/local_sgd.py
tushar00jain: no lol probably just slipped in because of other trial and errors

tushar00jain: reverted

torchft/collectives_test.py
tushar00jain: updated

torchft/work.py
tushar00jain: updated

torchft/collectives.py
```diff
@@ -391,7 +400,6 @@ def callback(fut: Future[list[torch.Tensor]]) -> list[torch.Tensor]:
                 fut.wait()
                 # Dequantize the result back to the original precision
                 fused_dequantize_from_fp8(tensors, quantized_tensors, world_size)
-                return tensors
 
-        fut = fut.then(callback)
-        return fut
+        fut.add_done_callback(callback)
+        return work
```
d4l3k, line 405: should this be using _WorkWrapper ?

tushar00jain: that's returned by manager. the manager calls this method inside allreduce

torchft/local_sgd.py
tushar00jain: it does't so fixed it

PR #248, opened 2025-07-26
option 2 - call worl.wait inside future callback

tushar00jain:


torchft/work.py
```diff
@@ -19,18 +19,3 @@ def wait(self, timeout: Optional[timedelta] = None) -> bool:
...
-class _WorkWrapper(dist._Work):
-    def __init__(
-        self, work: dist._Work, fut: torch.futures.Future[torch.Tensor]
-    ) -> None:
-        super().__init__()
-        self._work = work
-        self._fut = fut
-
-    def wait(self, timeout: Optional[timedelta] = None) -> bool:
-        return True
```
d4l3k, line 33: The original behavior here seems wrong -- wait should wait until the work is fully complete?

torchft/manager.py
```diff
@@ -932,3 +904,59 @@ def warn(self, msg: str) -> None:
...
+
+                return self._tensor
+
+        self._fut = self._fut.then(callback)
+        self._fut = self._manager.wrap_future(self._fut, self._tensor)
+
+        return True
+
+    def get_future(self) -> torch.futures.Future[torch.Tensor]:
+        self.wait()
```
d4l3k: .wait() should be a blocking call, we probably want to invert this logic and make .wait() call get_future() instead

torchft/work.py
tushar00jain: original implementation wasn't meant to work with gloo. it's changing in this diff to do exactly that. we also *want* wait to not block. maybe this overload should be for block_current_stream?

torchft/manager.py
tushar00jain: yep, the .wait() and .get_future() should only be called when we want to block. for diloco, it's called at the sync step. iiuc manager allreduce method doesn't get called from hsdp so it's unrelated there. this also made me realize we call .get_future() for bucketized allreduce though where we don't want to block. thinking we can pass the callback to manager allreduce for that. guess it's not nice api vise to block in this method. so maybe we need option 1 but it seems to have some issues

the .wait() already calls .get_future(), we want to make sure when users of this api call .get_future(), we've already setup the stream dependencies for work and the tensor division

torchft/work.py
tushar00jain: updated the orininal implementation

torchft/manager.py
```diff
@@ -938,3 +925,92 @@ def warn(self, msg: str) -> None:
...
+            torch.futures.Future[torch.Tensor], torch.futures.Future[None]
+        ] = work.get_future()
+
+        self._stream: Optional[torch.cuda.Stream] = (
+            torch.cuda.current_stream() if torch.cuda.is_available() else None
+        )
+
+        self._is_set_future_callback_called = False
+
+    def _set_future_callback(
```
d4l3k, line 947: Why does this need to live in the Work object? Can't we pass the stream + the future to the _WorkWrapper and have it manage things correctly?

tushar00jain: that would be ideal but doesn't work in all cases i think

- for nccl, we need to call work.wait() before doing everything in _set_future_callback(), otherwise stream dependency is not hooked up in the right order i think i.e. we could end up calling future.wait before work.wait
- for cpu, we can't call work.wait() because that'll block 

these 2 conflict with each other, so this is what i came up with

torchft/manager.py
tushar00jain: get_future now calls `block_current_stream`

torchft/manager.py
d4l3k: @tushar00jain you can use `work.synchronize()` to setup the dependency in a guaranteed non-blocking way

tushar00jain: @d4l3k that's for nccl right? for nccl you mentioned `block_current_stream` also just calls `synchronize` but it also works in a non-blocking way for gloo. we needed `block_current_stream` for that because i'm guessing `syncrhronize` does't do that for gloo.


also based on our discussion offline, the current api's work for all cases and have the same semantics as the underlying process group work

- in torchft, we only ever use `work.wait()`, and we call it only when we need to synchronize 
  - for nccl, and gloo with cuda, this sets up stream deps properly with a custom stream that we synchronize on to wait for the allreduce to finish along with the future associated with that work
  - for gloo with cpu, it just blocks until the work is done. the future callbacks run after the work is done
- that was a lie, we also call `work.get_future()` in bucketized allreduce
  - in this case we call `block_current_stream` first to set up stream dep for nccl (just a proxy to work.synchronize), and gloo with cuda. we also add a callback to the future chain but carefully set up the stream dep after all the other stream deps have been set up. that's why we call `block_current_stream` in get_future anyway
  - for gloo with cpu, it doesn't call anything on work because futures anyway run after the work is done
- we will call work.block_current_stream for hsdp in torchtitan -- this is pretty much the same as the case above for bucketized allreduce
- for ddp, we call get_future but don't expect users to do anything besides calling .wait on that future

In the future,

- we can consider creating our own future instead of using torch.futures.Future that sets up stream deps like the way we want it to
- consider simplifying the implementation of `_ManagedWork` (the above will also help us do that)

torchft/manager.py
```diff
@@ -954,3 +941,103 @@ def warn(self, msg: str) -> None:
...
+                return self._tensor
+
+        fut = self._fut
+        fut = fut.then(callback)
+        fut = self._manager.wrap_future(fut, self._tensor)
+        self._fut = fut
+
+        self._is_set_future_callback_called = True
+
+    def wait(self, timeout: Optional[timedelta] = None) -> bool:
```
d4l3k, line 979: .wait() should set a dependency between the work and the current stream -- it looks like we're running all operations on self._stream?

torchft/manager.py
```diff
@@ -954,3 +941,103 @@ def warn(self, msg: str) -> None:
...
+        with (
+            torch.cuda.stream(self._stream)
+            if self._stream is not None
+            else nullcontext()
+        ):
+            self._fut.wait()
+
+        return True
+
+    def block_current_stream(self, timeout: Optional[timedelta] = None) -> None:
```
d4l3k, line 998: we probably shouldn't rely on this until we've thought this through more / tested

tushar00jain: yeah we can test it more before we change the hsdp implementation. think we can also do some other alternative for bucketized allreduce and ddp without having to use block_current_stream

PR #249, opened 2025-07-29
fix managed pg allreduce

tushar00jain:
Summary:
managed pg allreduce should just call manager's allreduce

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -1062,30 +1062,6 @@ class FakeProcessGroupWrapper(ProcessGroupWrapper):
         return work
 
 
-class _ManagedWork(Work):
-    def __init__(self, manager: "Manager", work: Work, default_result: object) -> None:
-        super().__init__()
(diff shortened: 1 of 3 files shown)


torchft/process_group.py
```diff
@@ -1102,23 +1078,7 @@ def __init__(self, manager: "Manager") -> None:
...
-        except Exception as e:
-            self._manager.report_error(e)
-            return _DummyWork(tensors)
-
-        return _ManagedWork(
-            self._manager,
-            work,
-            tensors,
-        )
+        return self._manager.allreduce(tensors)
```
d4l3k: We should probably have a hybrid between these two options -- we should check the opts object for the ReduceOp and only call manager.allreduce for ReduceOp.AVG

With this change we'll silently always do AVG via manager

tushar00jain: updated manager to support both avg and sum

PR #250, opened 2025-07-29
Fix project urls to not point to PyTorch-labs

ZainRizvi:
Fixing old hang-overs from before this became a PyTorch org project


H-Huang (approved):
LGTM

PR #253, opened 2025-08-04
clean up managed work

tushar00jain:
Summary:
- improve abstraction and api contract offered by  managed work - users can change future return types, ensures users use correct cuda streams
- added a managed future that is lazily attaches callbacks - based on when users call wait on the work  or the future

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/torchft/pull/253).
* #249
* __->__ #253

### DIFF
--- a/torchft/ddp_test.py
+++ b/torchft/ddp_test.py
@@ -14,7 +14,7 @@ from torch.distributed.distributed_c10d import Work
 from torch.futures import Future
 
 from torchft.ddp import DistributedDataParallel, PureDistributedDataParallel
-from torchft.manager import Manager
+from torchft.manager import Manager, _ManagedWork
 from torchft.process_group import ProcessGroupBabyGloo, ProcessGroupGloo
 from torchft.work import _DummyWork
 
@@ -41,6 +41,7 @@ class TestDDP(TestCase):
 
         call_count = 0
 
+        # pyre-ignore[53]: Captured variable `manager` is not annotated.
         def allreduce(
             tensor: torch.Tensor,
         ) -> Work:
@@ -48,7 +49,8 @@ class TestDDP(TestCase):
 
             call_count += 1
 
-            return _DummyWork(tensor)
+            work = _DummyWork(tensor)
+            return _ManagedWork(manager, work, tensor)
 
         manager.allreduce = allreduce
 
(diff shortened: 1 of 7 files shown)


torchft/ddp_test.py
```diff
@@ -41,14 +41,16 @@ def test_ddp(self) -> None:
 
         call_count = 0
 
+        # pyre-ignore[53]: Captured variable `manager` is not annotated.
         def allreduce(
             tensor: torch.Tensor,
         ) -> Work:
-            nonlocal call_count
+            nonlocal call_count, manager
```
d4l3k: nit: we don't assign manager so shouldn't need to specify it here

torchft/manager.py
```diff
@@ -418,7 +419,20 @@ def allreduce(
...
                 work = self._pg.allreduce([tensor], ReduceOp.SUM)
 
-            return _ManagedWork(work, self, tensor, num_participants)
+            # schedule grad normalization as a continuation
+            # on the Future
+            @torch.profiler.record_function("torchft::manager::allreduce::callback")
+            def callback(
+                fut: torch.futures.Future[list[torch.Tensor]],
+            ) -> torch.Tensor:
+                nonlocal num_participants, tensor
```
d4l3k: nbit: since `/=` is a mutating operation we shouldn't need nonlocal?

torchft/manager.py
```diff
@@ -943,81 +957,251 @@ def exception(self, msg: str) -> None:
...
+    def add_done_callback(
+        self, callback: Callable[[torch.futures.Future[T]], None]
+    ) -> None:
+        raise RuntimeError("should not be called")
+
+    def set_result(self, result: T) -> None:
+        raise RuntimeError("should not be called")
+
+    def set_exception(self, result: T) -> None:
+        raise RuntimeError("should not be called")
```
d4l3k: nit: NotImplementedError instead of RuntimeError? also maybe we can make this more of a "this method is not supported" with an explanation. Could also make a custom class since this is used in so many spots

d4l3k (approved):
LGTM thanks for cleaning this up!

PR #259, opened 2025-08-20
Fixed the issue where quorum retry could not catch non-timeout errors.

zhengchenyu:
#228 implements quorum retry, a very important feature. This ensures that even if Lighthouse temporarily crashes, training will not be terminated.

I tested this by manually shutting down Lighthouse and then restarting it. I found that training could not be resumed. Found the following log:

```
2025-08-20T17:25:50.494 [INFO] [torchft::manager] - Got quorum response success: Err(Status { code: Unknown, message: "transport error", source: Some(tonic::transport::Error(Transport, hyper::Error(Io, Kind(ConnectionReset)))) })
```

The reason is that `tokio::time::timeout(timeout, client.quorum(request)).await` returns `Result(Result)`. When Lighthouse is shut down, it returns Ok(Err), so the Lighthouse client cannot be recreated, resulting in the inability to resume training.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -253,13 +253,40 @@ impl Manager {
             let result = tokio::time::timeout(timeout, client.quorum(request)).await;
 
             match result {
-                Ok(response) => {
-                    return response;
+                Ok(Ok(response)) => {
+                    info!("Got quorum response success: {:?}", response);
+                    return Ok(response);
+                }
+                Ok(Err(e)) => {
+                    info_with_replica!(
+                        self.replica_id,
+                        "lighthouse quorum failed. error: {}",
+                        e.to_string()
+                    );
+
+                    if retry_count == self.quorum_retries {
+                        return Err(Status::internal(format!(
+                            "lighthouse quorum failed after {} retries. error: {}",
+                            retry_count,
+                            e.to_string(),
+                        )));
+                    }
+
+                    // In general, quorum failure will return immediately,
+                    // but not waiting for the timeout.
+                    tokio::time::sleep(timeout).await;
+
+                    // Reset the client since the lighthouse server might have failed
+                    // If this also fails, consider increasing `connect_timeout`.
+                    let _ = self.create_lighthouse_client().await;
+
+                    retry_count += 1;
+                    info!("lighthouse quorum retried: {}", retry_count);
                 }
                 Err(e) => {
                     info_with_replica!(
                         self.replica_id,
-                        "lighthouse quorum failed. error: {}",
+                        "lighthouse quorum timeout. error: {}",
                         e.to_string()
                     );
 


src/manager.rs
```diff
@@ -253,10 +253,10 @@ impl Manager {
             let result = tokio::time::timeout(timeout, client.quorum(request)).await;
 
             match result {
-                Ok(response) => {
-                    return response;
+                Ok(Ok(response)) => {
+                    return Ok(response);
                 }
-                Err(e) => {
+                Ok(Err(e)) => {
```
tushar00jain, line 260: can we simply do this and have everything else unchanged?

### 
```suggestion
                _ => {
```

tushar00jain (changes requested):
LGTM except can be simplified a bit

src/manager.rs
zhengchenyu: @tushar00jain 
If I replicate `Ok(Err(e))` and `Err(e)` with `_`, e will be hidden and we won't be able to log the error message. 
Err(e) means quorum timeout, It mean that have wait for `timeout`.
Here `Ok(Err(e))` means quorum failed, usually return immediately, then delay `timeout` so that we can keep similar logic with quorum timeout. So I split two match. 
How about this? Or simply this without any error messge?

tushar00jain: I see, will this work `Ok(Err(e)) | Err(e) => {`?

zhengchenyu: @tushar00jain 
I tried, if use `Ok(Err(e1)) | Err(e2) => {`,  the type of e1 is `tonic::Status`, and the type of e2 is `Elapsed`, are different type, is not allowed by rust.
Or we can use two levels of matching, but it is also not concise.

In fact, I've tried many approaches before, but none of them were very concise. becuase I want to print different logs and provide different delays. In this example, I'm causing Lighthouse to crash, which will immediately return an OK (Error). I think 100ms of delay seems too short for the Lighthouse down.

tushar00jain: I see makes sense. I was trying to see if we can avoid the duplication. Maybe we can move the following after the match arm since we return early in cases where this code shouldn't be triggered. 

```
                    // In general, quorum failure will return immediately,
                    // but not waiting for the timeout.
                    tokio::time::sleep(timeout).await;

                    // Reset the client since the lighthouse server might have failed
                    // If this also fails, consider increasing `connect_timeout`.
                    let _ = self.create_lighthouse_client().await;

                    retry_count += 1;
```


Using `timeout` seems a bit risky (seems like an existing problem). Basically it means we should set `timeout * retries * 2` to be less than `quorum_timeout` in manager.py but `quorum_timeout == timeout` (that's how timeout is determined). I think we need to break the passed in timeout such that `x * retries * 2 == timeout` and use `x` for the timeouts used in this method.

zhengchenyu: Great suggestion! I have fix it. This is indeed more concise.
And `quorum_timeout` and `timeout` do need to be carefully configured, and perhaps a unified configuration is needed.

src/manager.rs
```diff
@@ -252,34 +252,45 @@ impl Manager {
...
-                    }
+            if retry_count == self.quorum_retries {
+                return Err(Status::internal(format!(
+                    "lighthouse quorum failed after {} retries.",
+                    retry_count,
+                )));
+            }
 
-                    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+            tokio::time::sleep(timeout).await;
```
tushar00jain: is this supposed to be `sleep_time`

src/manager.rs
tushar00jain: sure that can be addressed separately. if you're up for it, a unit test for this behavior will be nice

src/manager.rs
```diff
@@ -252,34 +252,45 @@ impl Manager {
 
             let result = tokio::time::timeout(timeout, client.quorum(request)).await;
 
+            let sleep_time;
```
tushar00jain: needs `mut` and type?

src/manager.rs
zhengchenyu: I have add new unit test which can reproduce the bug.
Thanks for your reminder and unit test, I just found that `quorum_timeout == timeout`, so we can not sleep `timeout`. The two timeout for quorum are indeed the same:

* (1) the `timeout` for grpc from manager client to manager server. (The quorum_timeout corresponding to your reminder)
* (2) the `timeout` for grpc from lighthouse client to lighthouse server. (The timeout corresponding to your reminder)

If I sleep this `timeout`, the manager client can not get result because of manager rpc client timeout. So I reset sleep time to `max(100, timeout /max(self.quorum_retries + 1, 1))`. By this way, the lighthouse client can retry the quorum `quorum_retries` times.

And sorry I do not understand you said the 2 in`timeout * retries * 2`. I do not know why there is 2.

tushar00jain: we also have a `timeout` on the connection to lighthouse for each retry. and each retry also waits `timeout` if the lighthouse connection fails. so combining both, `retries * timeout + retries * timeout` in the worst case.

Issue #257, opened 2025-08-18
[RFC] [Intel GPU] Extending TorchFT to Support Intel GPU with XCCL Backend


siju-samuel:
## Abstract
We propose extending TorchFT to support [Intel XPU](https://docs.pytorch.org/docs/stable/notes/get_start_xpu.html) devices by adapting ProcessGroups to use the "xccl" backend. This extension will enable Intel GPU users to benefit from the same per-step fault tolerance guarantees currently available on CUDA, including checkpointing, elastic recovery, node failure detection, and AFT-style recovery.

## Motivation
TorchFT today provides per-step fault tolerance for PyTorch training, allowing jobs to continue without restarting the entire loop when failures occur. It currently supports CUDA backends (NCCL/Gloo) for FT-DDP, FSDP/HSDP, and checkpointing.

[Intel’s XPU](https://docs.pytorch.org/docs/stable/notes/get_start_xpu.html) (Supported via `torch.xpu` since PyTorch 2.5) is gaining adoption in large-scale training and inference. To fully leverage TorchFT on Intel GPUs, we need native XPU integration with the "xccl" backend. Without this, Intel GPU users cannot take advantage of TorchFT’s core features (fault-tolerant DDP/FSDP, elastic scaling, checkpointing).

Adding XPU support ensures parity with CUDA workflows and expands TorchFT’s utility to a broader hardware ecosystem. The architecture diagram below illustrates TorchFT’s high-level design: (Added XCCL to the Original Diagram)

<img width="1663" height="460" alt="Image" src="https://github.com/user-attachments/assets/ee165aac-c86a-4ef1-98f0-be9a38d8418b" />

### API and Code Changes

1. **Device Checks**
   - Replace CUDA-specific checks with device-agnostic APIs where possible.
   - Example:
     ```python
     # Instead of current CUDA-only
     torch.cuda.is_available()
     torch.cuda.set_device(id)

     # Use device-agnostic APIs for uniform detection across CUDA and XPU
     torch.accelerator.is_available()
     torch.accelerator.set_device(id)
     ```

2. **Streams & Events**
   - Replace `torch.cuda.Stream` / `torch.cuda.Event` with device-agnostic `torch.Stream` / `torch.Event`.

3. **Process Groups**
   - Extend TorchFT’s ProcessGroup wrappers to support:
     - `ProcessGroupXCCL`
     - `ProcessGroupBabyXCCL`

4. **DDP/FSDP/HSDP Wrappers for Tests**
[...]

PR #260, opened 2025-08-20
[Intel GPU] Extending TorchFT to Support Intel GPU with XCCL Backend

siju-samuel:
This PR introduces the foundational support for enabling XPU devices with XCCL as the backend in TorchFT as per [RFC](https://github.com/pytorch/torchft/issues/257)

Key highlights:

- Added `ProcessGroupXCCL/ProcessGroupBabyXCCL`  implementation
- Integrated XPU device handling into manager/processgroup
- Managed streams and events
- Updated example train_ddp.py to run in Intel GPU environment

cc @tushar00jain @jeromean @gujinghui @zhangxiaoli73 @rbabukv

### DIFF
--- a/torchft/futures.py
+++ b/torchft/futures.py
@@ -161,12 +161,21 @@ class _TimeoutManager:
             handle,
         )
 
-        stream: Optional[torch.cuda.Stream] = (
-            torch.cuda.current_stream() if torch.cuda.is_available() else None
+        stream: Optional[torch.Stream] = (
+            torch.accelerator.current_stream() if torch.accelerator.is_available() else None
         )
 
         def callback(fut: Future[T]) -> None:
-            with torch.cuda.stream(stream) if stream is not None else nullcontext():
+            if stream is not None:
+                if torch.cuda.is_available():
+                    context = torch.cuda.stream(stream)
+                elif torch.xpu.is_available():
+                    context = torch.xpu.stream(stream)
+                else:
+                    context = nullcontext()
+            else:
+                context = nullcontext()
+            with context:
                 handle.cancel()
                 try:
                     timed_fut.set_result(fut.wait())
@@ -186,7 +195,7 @@ class _TimeoutManager:
 
         loop = self._maybe_start_event_loop()
 
-        event: torch.cuda.Event = torch.cuda.Event()
+        event: torch.Event = torch.Event()
         event.record()
 
         def handler() -> None:
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -276,12 +276,13 @@ class Manager:
         self._pg = pg
         self._manager: Optional[ManagerServer] = None
 
-        self._recovery_stream: Optional["torch.cuda.Stream"] = (
-            torch.cuda.Stream() if torch.cuda.is_available() else None
-        )
+        if torch.accelerator.is_available():
+            self._recovery_stream: Optional["torch.Stream"] = torch.Stream()
+        else:
+            self._recovery_stream = None
 
         # Used to synchronize recovery operation
-        self._recovery_event: Optional[torch.cuda.Event] = None
+        self._recovery_event: Optional[torch.Event] = None
 
         if self._group_rank == 0:
             if port is None:
@@ -413,8 +414,13 @@ class Manager:
             # Run the allreduce async and save the work object so we can wait on
             # it later.
             if should_quantize and IS_TRITON_AVAILABLE:
+                if torch.accelerator.is_available():
+                    current_stream = torch.accelerator.current_stream()
+                else:
+                    current_stream = None
+
                 work = allreduce_quantized(
-                    [tensor], ReduceOp.SUM, self._pg, torch.cuda.current_stream()
+                    [tensor], ReduceOp.SUM, self._pg, current_stream
                 )
             else:
                 work = self._pg.allreduce([tensor], ReduceOp.SUM)
@@ -482,9 +488,10 @@ class Manager:
 
         fut = future_timeout(fut, timeout or self._timeout)
 
-        stream: Optional[torch.cuda.Stream] = (
-            torch.cuda.current_stream() if torch.cuda.is_available() else None
-        )
+        if torch.accelerator.is_available():
+            stream = torch.accelerator.current_stream()
+        else:
+            stream = None
 
         # schedule error handling as a continuation on the Future
         def callback(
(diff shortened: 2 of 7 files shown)


torchft/futures.py
```diff
@@ -161,12 +161,21 @@ def register(self, fut: Future[T], timeout: timedelta) -> Future[T]:
...
 
-        stream: Optional[torch.cuda.Stream] = (
-            torch.cuda.current_stream() if torch.cuda.is_available() else None
+        stream: Optional[torch.Stream] = (
+            torch.accelerator.current_stream() if torch.accelerator.is_available() else None
         )
 
         def callback(fut: Future[T]) -> None:
-            with torch.cuda.stream(stream) if stream is not None else nullcontext():
+            if stream is not None:
```
tushar00jain: also extract this into a separate method that can be reused in other places? it'll also make the code changes minimal

torchft/manager.py
```diff
@@ -276,12 +276,13 @@ def __init__(
         self._pg = pg
         self._manager: Optional[ManagerServer] = None
 
-        self._recovery_stream: Optional["torch.cuda.Stream"] = (
-            torch.cuda.Stream() if torch.cuda.is_available() else None
-        )
+        if torch.accelerator.is_available():
```
tushar00jain: i'd just keep the previous structure i.e. removing `cuda` from `torch.cuda.Stream` so to keep the code changes minimal

torchft/manager.py
```diff
@@ -482,17 +488,27 @@ def wrap_future(
 
         fut = future_timeout(fut, timeout or self._timeout)
 
-        stream: Optional[torch.cuda.Stream] = (
-            torch.cuda.current_stream() if torch.cuda.is_available() else None
-        )
+        if torch.accelerator.is_available():
```
tushar00jain: same here, let's keep the code changes minimal

torchft/manager.py
```diff
@@ -546,7 +562,7 @@ def start_quorum(
             shrink_only=shrink_only,
             quorum_timeout=timeout or self._quorum_timeout,
             curr_device=(
-                torch.cuda.current_device() if torch.cuda.is_available() else -1
+                torch.accelerator.current_device_index() if torch.accelerator.is_available() else -1
```
tushar00jain: why did this change from `current_device` to `current_devide_index`?

train_ddp.py
```diff
@@ -75,19 +91,20 @@ def state_dict():
...
+        pg = ProcessGroupXCCL(timeout=timedelta(seconds=30))
+    else:
+        device = "cpu"
+        pg = ProcessGroupGloo(timeout=timedelta(seconds=5))
 
     transport = PGTransport(
         pg,
         timeout=timedelta(seconds=10),
-        device=("cuda" if torch.cuda.is_available() else "cpu"),
+        device=device,
```
tushar00jain: the device here can be different from the device for the process group. so one could use cpu gloo for process group but use nccl/xccl here

train_ddp.py
```diff
@@ -36,6 +41,17 @@
 
 @record
 def main() -> None:
+    # Parse command line arguments
+    parser = argparse.ArgumentParser(description='TorchFT DDP Training')
```
tushar00jain: does xccl not support profiling? 

seems like a reasonable change though would be good to put these changes in a separate pr

torchft/process_group.py
```diff
@@ -794,6 +798,183 @@ def errored(self) -> Optional[Exception]:
     def getBackendName(self) -> str:
         return "torchft-nccl"
 
+class _WorkXPUTimeout(Work):
```
tushar00jain: rename `_WorkCudaTimeout` to `_WorkAcceleratorTimeout` instead of duplicating?

torchft/futures.py
siju-samuel: Done

torchft/manager.py
siju-samuel: Done

torchft/manager.py
siju-samuel: `torch.accelerator.current_device` does not exist.
`torch.accelerator.current_accelerator` returns the device type (e.g., `device(type='cuda')`), whereas `current_device` would return the device index (e.g., `0`).
Since we need just the index, `torch.accelerator.current_device_index` is used instead.
Similarly, instead of using `cuda.set_device`, I used `accelerator.set_device_index` for consistency.

train_ddp.py
siju-samuel: Yeah. Added this code seperatly

train_ddp.py
siju-samuel: XCCL returns some error during profiling which we are looking into.
Will raise another PR for this change.

torchft/process_group.py
siju-samuel: Done

torchft/manager.py
siju-samuel: Done

tushar00jain:
@siju-samuel thanks for putting this together! The before/after for the profiles look good. What machine are you taking the profiles from and do we need any special machines to test out XCCL? Is there XCCL support in the pytorch profiler? The profiler doesn't seem to show the allreduce.

Some of the tests you mentioned that are failing in the CI are configured to be skipped so not sure why they're running.

siju-samuel:
> Do we need any special machines to test out XCCL? 

Yes, this is supported only with Intel GPU (XPU) hardware

> Is there XCCL support in the pytorch profiler? The profiler doesn't seem to show the allreduce.

Yeah. we support, but the trace collection method at kernel level is slightly different.  In the profile data i have given, the allreduce kernel came along with other kernels in same stream. Need to check whether there is actual issue in XPU or only tracing problem.

> CI Failure due to TypeError: cannot pickle 'torch.Event' object

Is fixed currently with this commit  https://github.com/pytorch/torchft/pull/260/commits/c65eed852bc73fd2425fd79d72df37db51f4504d

tushar00jain:
@siju-samuel for the mac unit tests, maybe the skipif is not working with parameterized expand. we can add a return statement explicitly inside the test if the platform is darwin. that should make the mac unit tests pass.

siju-samuel:
Thanks @tushar00jain. 
You are right. the `skipIf `was not working with parameterized. (only pytest `skipIf `works with parameterized).
Submitted a patch to skip inside function.

siju-samuel:
```
torchft/process_group_test.py::BabyNcclMultiPgTest::test_collective_08_reduce_scatter FAILED [ 78%]
        expected_sum = expected_row * world_sz
>       torch.testing.assert_close(out, expected_sum)
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 2 / 2 (100.0%)
E       Greatest absolute difference: 0.0 at index (0,) (up to 1e-05 allowed)
E       Greatest relative difference: 0.0 at index (0,) (up to 1.3e-06 allowed)
```

Looks like a random failure. CUDA tests were passing before.
MAC tests passed in last run.
@tushar00jain  Could you please trigger CI once again. Thanks

tushar00jain:
@siju-samuel I can trigger but haven't seen those failures before. Usually it's an indidication that something is wrong. Can you run a couple times locally?

tushar00jain (approved):

PR #264, opened 2025-09-11
Add monarch+torchft distributed training example

amirafzali:
README should provide all necessary context for the script. We plan to iterate from here to add the following:

1. Cleanup of allocator logic once it is upstreamed in Monarch
2. Support for hot spares / host-level supervision
3. We could consider adding TrainerActor and a fleshed out ReplicaActor (with host supervision) to the library once they are in a more complete state, so that they are easily accessible to users.

### DIFF
--- a/examples/monarch/README.md
+++ b/examples/monarch/README.md
@@ -1,0 +1,50 @@
+Monarch-TorchFT-TorchTitan Distributed Training Orchestrator
+
+This script orchestrates fault-tolerant distributed training using TorchTitan and TorchMonarch
+frameworks. It manages multiple training replicas across SLURM-scheduled compute nodes
+with automatic failure recovery and TorchFT lighthouse coordination.
+
+PREREQUISITES:
+- Access to a SLURM cluster with GPU nodes
+- Environment with nightly TorchFT, TorchTitan, and Monarch libraries installed.
+- TorchTitan training configuration file in script directory (debug_model.toml)
+- A training dataset (c4_test) and tokenizer in script directory
+
+CONFIGURATION:
+Before running, update the cluster-specific constants:
+- MACHINE: TorchX named resource for your cluster (currently: "gpu.xlarge")
+- MACHINE_MEMORY: Memory per machine in MB (currently: 2062607)
+You can also override the resource configuration manually:
+- https://docs.pytorch.org/torchx/main/specs.html#resource
+
+USAGE:
+    python train_distributed.py --help
+
+    Basic usage with 2 replicas, each with 1 node and 8 GPUs:
+        python train_distributed.py
+
+    Custom configuration:
+        python train_distributed.py --replica-count 3 --gpu-per-node 8 \
+            --host-per-replica 2 --training-steps 100
+
+    With remote TorchFT lighthouse:
+        python train_distributed.py --remote-lighthouse
+
+KEY COMPONENTS:
+- LighthouseActor: Coordination server for fault tolerance
+- TrainingActor: Individual trainer processes
+- ReplicaActor: Manages groups of trainers
+- OrchestrationManager: Top-level orchestration and failure recovery
+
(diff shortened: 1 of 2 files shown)


examples/monarch/README.md
```diff
@@ -0,0 +1,50 @@
+Monarch-TorchFT-TorchTitan Distributed Training Orchestrator
```
d4l3k: nit: Can we use markdown formatting for the title and different sections?

examples/monarch/train_distributed.py
```diff
@@ -0,0 +1,420 @@
...
+
+# ==== allocation boilerplate ====
+
+class LighthouseActor(Actor):
+    def __init__(self) -> None:
+        self.lighthouse = None
+
+    @endpoint
+    def start_lighthouse(self) -> str:
+        from torchft.coordination import LighthouseServer
```
colin2328, line 94: any reason for this inline import?

examples/monarch/train_distributed.py
```diff
@@ -0,0 +1,420 @@
...
+
+# delay before re-creating proc mesh on existing job. change as needed.
+PROC_ATTEMPT_DELAY = 10
+# proc attempts before getting a new scheduler allocation. change as needed.
+PROC_ATTEMPTS = 2
+# attempts before failing training on replica. change as needed.
+MAX_ATTEMPT = PROC_ATTEMPTS * 2
+
+
+class OrchestrationManager:
```
colin2328, line 209: I would probably add docstrings to these classes

examples/monarch/train_distributed.py
```diff
@@ -0,0 +1,420 @@
...
+from monarch.tools.components import hyperactor
+from monarch.tools.config import Config
+from monarch.utils import setup_env_for_distributed
+
+from torchtitan.config import ConfigManager, JobConfig
+from torchtitan.tools.logging import init_logger, logger
+from torchtitan.train import Trainer
+
+
+# ==== Allocation boilerplate - much of this will be upstreamed into Monarch ====
```
colin2328, line 29: what are you thinking will be upstreamted into monarch?

colin2328: cc @kiukchung  any suggestions on OSS example here?

examples/monarch/train_distributed.py
```diff
@@ -0,0 +1,420 @@
...
+async def main() -> None:
+    init_logger()
+
+    args = parse_args()
+    job_spec = make_job_spec(args)
+
+    atexit.register(MonarchSlurm.kill_jobs)
+    orchestrator = OrchestrationManager(job_spec)
+    try:
+        await orchestrator.start_lighthouse()
```
colin2328, line 420: do we need to fully await start_lighthouse before start_training?

examples/monarch/train_distributed.py
amirafzali: yes, I opened this bug: https://github.com/meta-pytorch/monarch/issues/804
though i haven't tried in a few week, maybe somebody fixed it

examples/monarch/train_distributed.py
amirafzali: I talked with Kiuk about this a little.
1. The pattern where we have to manually manage the job handles is not desirable, there are ways we can handle this on the library side
2. IMO, the user should not have to write out the logic to get a proc mesh allocator for slurm. There should be an API that provides this

examples/monarch/train_distributed.py
amirafzali: yes, it only takes a few seconds at the beginning of the script and is a dependency for replica coordination

examples/monarch/train_distributed.py
colin2328: add a comment here? (you can link ot the issue)

PR #270, opened 2025-09-25
add failure injector for monarch training script

amirafzali:
This introduces a failure injector with 5 failure modes:
- SEGFAULT: Triggers a SIGSEGV on the process
- DEADLOCK = Deadlocks the GIL, resulting in ProcessGroupNCCL timeout and terminal failure
- KILL_PROC: Immediately kills the process with non-zero exit code
- COMMS = Forcefully aborts the ProcessGroup and NCCL communicator
- KILL_SLURM = Kills a random replica SLURM job

It can be enabled with the flag `--with--failure`, and it runs async every 60 seconds.

### DIFF
--- a/examples/monarch/train_distributed.py
+++ b/examples/monarch/train_distributed.py
@@ -24,46 +24,53 @@ from monarch.utils import setup_env_for_distributed
 from torchtitan.config import ConfigManager, JobConfig
 from torchtitan.tools.logging import init_logger, logger
 from torchtitan.train import Trainer
+from utils.failure import Failure, FailureActor, FailureController
 
 
 # ==== Allocation boilerplate - much of this will be upstreamed into Monarch ====
 class MonarchSlurm:
     # Cluster Configuration - update these values for your specific cluster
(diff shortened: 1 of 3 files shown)


examples/monarch/train_distributed.py
```diff
@@ -197,13 +208,29 @@ async def start_replica(self) -> None:
...
 
+    @endpoint
+    async def inject_failure(self, failure_type: Failure):
+        if self.failure_actors:
+            try:
+                logger.info(
+                    f"{self.uid} Injecting failure ({failure_type}) into random trainer"
+                )
+
+                await self.failure_actors.fail.choose(failure_type)
```
d4l3k, line 217: .choose picks an arbitrary training?

amirafzali: > .choose picks an arbitrary training?

yup, choose will send it to one random trainer in the replica mesh

colin2328 (approved):
LGTM! thanks

PR #292, opened 2025-11-18
Keep the training data continuous and the total batch size constant regardless of changes in the replica world size.

zhengchenyu:
The current `train_ddp.py` has two problems:
* It cannot guarantee the sequential reading of each sample. For example, the replica group world size is 3, but only 2 replicas are working. Some samples will be missing.
* When the replica group world size changes, the total batch size used for gradient aggregation will change. This makes idempotency computation impossible.

The following modifications were made:
* `SkipDistributedSampler` is provided to ensure that training can resume from any offset.
* The dataloader is reconfigured when the quorum changes.
* For the training rounds that were just initialized and when the quorum changed, the commit will be abandoned due to the setting of the dirty flag.
* Add example `train_ddp2.py`.

### DIFF
--- a/train_fsdp2_fix_batch.py
+++ b/train_fsdp2_fix_batch.py
@@ -1,0 +1,317 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import hashlib
+import logging
+import os
+import sys
+from datetime import timedelta
+from itertools import chain
+
+os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
+REPLICA_GROUP_ID = int(os.environ.get("REPLICA_GROUP_ID", 0))
+os.environ["NCCL_HOSTID"] = str(REPLICA_GROUP_ID)
+
+import torch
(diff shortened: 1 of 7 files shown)


zhengchenyu:
I add sequence diagram to illustrate the start of training and the scaling up from 1 replica world size to 2. local batch = 2, total batch = 4. Each color represents one iteration in the while training process.
The yellow iteration is the start of training . The orange iteration is normal training when replica world size is 1. The green iteration is scaling up from 1 replica world size to 2. The blue iteration is normal training when replica world size is 2.

image

train_fsdp2_fix_batch.py
```diff
@@ -0,0 +1,317 @@
...
+                loss = criterion(out, labels)
+                # For fsdp2, synchronization must be performed every time; `no_sync` cannot be used.
+                # This is because if `all_reduce_hook` is executed in `_fsdp_collectives.py`, it applies
+                # to the temporary `reduce_output` variable. If synchronization is only performed in the
+                # last step, each shard will lose the gradients of the previous `accumulation_steps - 1`
+                # steps from other shards.
+                loss.backward()
+                total_loss += loss.item()
+
+            if accumulation_steps > 1:
```
zhengchenyu, line 256: TODO: use `loss * (1 / accumulation_steps)`?

PR #315, opened 2026-02-24
Cleanup deprecated usages of package monarch.utils in torchft

dulinriley:
Summary:
Package `monarch.utils` was deprecated in 0.2 and removed in 0.4.
Migrate to the API it forwarded to.

Differential Revision: D93802250

### DIFF
--- a/examples/monarch/train_distributed.py
+++ b/examples/monarch/train_distributed.py
@@ -16,7 +16,7 @@ from typing import Dict
 import torch
 from monarch.actor import Actor, current_rank, endpoint, HostMesh, ProcMesh, this_host
 from monarch.job import SlurmJob
-from monarch.utils import setup_env_for_distributed
+from monarch.spmd import setup_torch_elastic_env_async
 from torchtitan.config import ConfigManager, JobConfig
 from torchtitan.tools.logging import init_logger, logger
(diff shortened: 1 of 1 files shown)


allenwang28 (approved):
Review automatically exported from Phabricator review in Meta.

JiwaniZakir (commented):
The rename from `setup_env_for_distributed` to `setup_torch_elastic_env_async` in `train_distributed.py` line 160 is straightforward, but it's worth verifying that the two functions are behaviorally equivalent and not just nominally similar. The new name encodes more specificity — "torch_elastic" implies it's scoped to Torch Elastic env vars — so if `setup_env_for_distributed` previously set up a broader set of environment variables beyond Torch Elastic, this swap could silently drop some initialization. It would help to include a brief comment or reference to the `monarch` changelog/migration guide confirming this is a pure rename. Additionally, since the `async` suffix is now explicit in the function name, it's worth confirming the call site's `await` was already present (it is, on line 160), which is good — this is consistent. A quick integration test run or a note in the PR description pointing to upstream `monarch` source confirming functional equivalence would make this easier to validate.

dulinriley:
`setup_env_for_distributed` just called `setup_torch_elastic_env_async` internally, so they are the same.

PR #319, opened 2026-03-24
Add torchcomms ProcessGroup shim for fault-tolerant reconfiguration

d4l3k:
## Summary
- Adds `ProcessGroupTorchComms` which wraps a `torchcomms.TorchComm` communicator as a `torchft.process_group.ProcessGroup`
- Translates all c10d ProcessGroup collective operations (allreduce, allgather, broadcast, reduce_scatter, alltoall, barrier, send, recv) to the torchcomms API

## Test plan

Added basic unit tests and E2E integration test using mccl.

```
pytest torchft/torchcomms_test.py
pytest torchft/manager_integ_test.py -k torchcomms
```

### DIFF
--- a/torchft/torchcomms.py
+++ b/torchft/torchcomms.py
@@ -1,0 +1,324 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+TorchComms Process Group
+=========================
+
+This module provides a shim layer that wraps a ``torchcomms.TorchComm``
+communicator as a ``torchft.process_group.ProcessGroup``, enabling
+fault-tolerant reconfiguration via the TorchComm ``reconfigure()`` API.
+
+On each :meth:`configure` call, init-handles are exchanged via the store
+and ``TorchComm.reconfigure()`` is called.  The same ``TorchComm`` object
+is reused across reconfigurations.
+"""
+
+import logging
+from datetime import timedelta
+from typing import List, Optional, Union
+
+import torch
+
+# pyre-fixme[21]: Could not find a module corresponding to import `torchcomms`.
+import torchcomms
+from torch.distributed.distributed_c10d import (
+    AllgatherOptions,
+    AllreduceCoalescedOptions,
+    AllreduceOptions,
+    AllToAllOptions,
+    BarrierOptions,
+    BroadcastOptions,
+    ReduceOp,
+    ReduceScatterOptions,
+    Work,
+)
+from torchft.process_group import create_store_client, ProcessGroup
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class _TorchCommsWork(Work):
+    """
+    Wraps a ``torchcomms.TorchWork`` to satisfy the c10d ``Work`` interface.
+
+    Warning: this class only supports futures partially. CPU futures require the
+    user to call wait() for the future to resolve. GPU futures are automatically
+    resolved to match stream semantics.
+    """
+
+    def __init__(
+        self,
+        device: torch.device,
+        # pyre-fixme[11]: Annotation `TorchWork` is not defined as a type.
+        work: torchcomms.TorchWork | None = None,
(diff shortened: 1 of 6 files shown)


torchft/torchcomms.py
```diff
@@ -0,0 +1,372 @@
...
+    ReduceOp,
+    ReduceScatterOptions,
+    Work,
+)
+from torchft.process_group import create_store_client, ProcessGroup
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class _TorchCommsWork(Work):
```
tushar00jain, line 44: this doesn't seem enough, we're using a few more api's like `get_future`. I had those implemented in https://github.com/meta-pytorch/torchft/pull/290/changes. or is this only meant for ddp?

torchft/torchcomms.py
```diff
@@ -0,0 +1,372 @@
...
+    def allreduce_coalesced(
+        self,
+        tensors: List[torch.Tensor],
+        opts: AllreduceCoalescedOptions,
+    ) -> Work:
+        op = _to_torchcomms_reduce_op(opts)
+        work = None
+        for t in tensors:
+            work = self.comm.all_reduce(t, op, async_op=True)
+        return _TorchCommsWork(work)
```
tushar00jain: this also needs things like `_run_context` and `_wrap_work`?

torchft/torchcomms_test.py
```diff
@@ -0,0 +1,329 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
tushar00jain, line 1: let's also test with the integration tests to see if it works with actual training?

tushar00jain (changes requested):

torchft/torchcomms.py
d4l3k: done, though it's a little hacky

torchft/torchcomms.py
```diff
@@ -0,0 +1,572 @@
...
+        self._fut: torch.futures.Future[object] = torch.futures.Future(
+            devices=[] if is_cpu else [device]
+        )
+        self._value = value
+        # Resolve the future immediately so that DDP's
+        # get_future().then(callback) pattern works without calling wait().
+        self._fut.set_result(self._value)
+
+    def wait(self, timeout: Optional[timedelta] = None) -> bool:
+        if self._work is not None:
```
tushar00jain, line 73: we also need to handle the time out here 

```
        with self._stream_timeout(self._comm, async_timeout):
            if self._work is not None:
                self._work.wait()

            # Always use cuda stream for timeout to avoid ProcessGroupNCCL
            # watchdog firing and crashing the process.
            if timeout is not None:
                torch.cuda.synchronize()
```

torchft/torchcomms.py
```diff
@@ -0,0 +1,572 @@
...
+        value: object = None,
+    ) -> Work:
+        """Wrap a ``torchcomms.TorchWork`` into a c10d ``Work``.
+
+        Subclasses can override this to add timeout handling or error tracking.
+        """
+        return _TorchCommsWork(self._device, work, value)
+
+    @contextmanager
+    def _run_context(self) -> Generator[None, None, None]:
```
tushar00jain: This is defined for PG NCCL, probably need separate sub class implementations just like PG

```
        timeout: timedelta = self._timeout

        def callback() -> None:
            logger.error(f"aborting after {timeout}!")
            self.abort()

        # when running in blocking mode we need to make sure collectives can
        # timeout
        with context_timeout(callback, timeout):
            yield
```

tushar00jain (changes requested):

torchft/torchcomms.py
d4l3k: I think this is actually fine since MCCL handles timeouts internally

d4l3k (commented):
updated with E2E testing with mccl, and dropped the ReconfigurableTorchComm wrapper since it was hacky and unreliable

tushar00jain (approved):

torchft/torchcomms.py
tushar00jain: let's document that it's only supposed to work with mccl

PR #324, opened 2026-05-06
Fix ProcessGroupGloo crash when setting unsupported backend options

wanna-01:
### Title

Fix `ProcessGroupGloo` crash when setting unsupported backend options

### Summary

This MR fixes a runtime crash in `ProcessGroupGloo._create_pg()` caused by unconditional assignment to Gloo backend options that are not present in PyTorch's `ProcessGroupGloo._Options`.

### Why this change is needed

Current code in `ProcessGroupGloo._create_pg()` does:

- `backend_class.options.global_ranks_in_group = ...`
- `backend_class.options.group_name = ...`

For Gloo, `backend_class.options` is `torch._C._distributed_c10d._Options`, which does not expose these attributes (it typically only has `backend`). Entering quorum on Gloo can therefore raise `AttributeError` and fail immediately.

NCCL path is not affected because NCCL `Options` does provide these fields.

### Root cause

The code assumes Gloo options support the same attributes as NCCL options.

### What this MR changes

In `torchft/process_group.py` (`ProcessGroupGloo._create_pg`):

- wrap assignment of `global_ranks_in_group` in `try/except AttributeError`
- wrap assignment of `group_name` in `try/except AttributeError`

So unsupported optional fields are skipped on Gloo instead of crashing.

### Behavioral impact

- prevents Gloo quorum initialization from hard-failing with `AttributeError`
- no behavior change for NCCL path
- keeps compatibility across PyTorch versions where Gloo options surface differs

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -653,9 +653,15 @@ class ProcessGroupGloo(ProcessGroupWrapper):
         backend_class._set_sequence_number_for_group()
 
         if self._global_ranks:
-            backend_class.options.global_ranks_in_group = self._global_ranks
+            try:
+                backend_class.options.global_ranks_in_group = self._global_ranks
+            except AttributeError:
+                pass
         if self._group_rank and self._group_world_size:
-            backend_class.options.group_name = f"torchft_quorum_{self._quorum_id}_rank_{self._group_rank % self._group_world_size}"
+            try:
+                backend_class.options.group_name = f"torchft_quorum_{self._quorum_id}_rank_{self._group_rank % self._group_world_size}"
+            except AttributeError:
+                pass
 
         pg._register_backend(
             torch.device("cpu"), ProcessGroup.BackendType.GLOO, backend_class


torchft/process_group.py
```diff
@@ -653,9 +653,15 @@ def _create_pg(self, store: Store, rank: int, world_size: int) -> BaseProcessGro
         backend_class._set_sequence_number_for_group()
 
         if self._global_ranks:
-            backend_class.options.global_ranks_in_group = self._global_ranks
+            try:
+                backend_class.options.global_ranks_in_group = self._global_ranks
```
tushar00jain, line 657: this is weird, the options for all the backends are inherited from https://github.com/pytorch/pytorch/blob/main/torch/csrc/distributed/c10d/Backend.hpp#L54 so this should be present. also this is needed for flight recorder to work properly (knowing what ranks are in the quorum)

tushar00jain (changes requested):

wanna-01:
Thanks for the review and for pointing this out. @tushar00jain 

I re-tested the issue with `torchft-nightly==2026.3.21`. The original failure was observed with a PyTorch 2.8.0a0-based environment, but the exact PyTorch commit/build was not recorded.

With the same torchft nightly on a current PyTorch build, `ProcessGroupGloo._Options` exposes both `global_ranks_in_group` and `group_name`, and the assignments succeed. The PyTorch v2.8.0 source also defines and binds these fields.

This suggests a compatibility issue with the specific PyTorch build used in the original environment, rather than a general lack of support in Gloo. The `try/except` workaround in this PR could silently drop metadata needed by the flight recorder, so I agree it is not an appropriate general fix.

I am closing this PR for now. If the issue reproduces on a supported PyTorch release, I can submit a new PR with the exact PyTorch version/commit and a focused compatibility fix.

PR #326, opened 2026-05-15
Add xpu support to the Unit Tests

Shreyashwa8:
Test script changes for #327 (landed as 05ec72e), per that PR's test plan. Adds XPU coverage across the unit and integration test suites using torch.accelerator, so the same tests run on CUDA and XPU.

 Also adds torchft/utils_test.py covering get_stream_context / record_event / synchronize from #260.
One non-test change: _maybe_share_tensors now only calls share_memory_() on CPU tensors — device tensors are already shared, and calling it on an XPU tensor fails. Required for pg_transport_test.py to pass on XPU. Validated on Intel Arc Pro B60 Graphics.

### DIFF
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -43,10 +43,12 @@ from torchft.process_group import (
     ProcessGroup,
     ProcessGroupBabyGloo,
     ProcessGroupBabyNCCL,
+    ProcessGroupBabyXCCL,
     ProcessGroupDummy,
     ProcessGroupGloo,
     ProcessGroupNCCL,
     ProcessGroupWrapper,
+    ProcessGroupXCCL,
 )
 from torchft.work import _DummyWork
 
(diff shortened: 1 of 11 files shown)


torchft/process_group_test.py
```diff
@@ -909,8 +995,10 @@ def _run_parallel(self, collective: str, device: str = "cpu") -> None:
         for rank in range(self.WORLD_SIZE):
             pg = self.pg_pool[rank]
             # Each worker calls `func(pg=pg, rank=rank, tensor=tensor, *args, **kwargs)`
-            if "cuda" in device:
+            if device == "cuda":
```
siju-samuel: `            if "cuda" in device: ` is correct i guess. what is device is cuda:0
same issue for xpu

torchft/process_group_test.py
```diff
@@ -720,6 +767,41 @@ def test_baby_nccl_apis(self) -> None:
         with self.assertRaisesRegex(OSError, "handle is closed"):
             a.allreduce([t], AllreduceOptions()).wait()
 
+    # pyre-fixme[56]: Pyre was not able to infer the type of argument
+    @skipUnless(torch.xpu.is_available(), "needs XPU")
+    @skipIf(True, "PyTorch XPU multiprocessing reductions not yet supported - see https://github.com/pytorch/pytorch/issues")
```
siju-samuel: wrong github link. which issue?

PR #327, opened 2026-05-26
Add Intel XPU (XCCL) support across torchft 

siju-samuel:
## Summary

Continuation of [PR-260](https://github.com/meta-pytorch/torchft/pull/260) as per [RFC-257](https://github.com/meta-pytorch/torchft/issues/257)

Extends torchft to run on Intel XPU accelerators alongside CUDA, using PyTorch's
device-agnostic accelerator API (`torch.accelerator.*`) so the same code path
covers both backends.

## Validation
Validated on Intel(R) Arc(TM) Pro B60 Graphics

## Test plan
Test script changes will be done as part of [PR-326](https://github.com/meta-pytorch/torchft/pull/326/)

CC: @tushar00jain @d4l3k  @rbabukv

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -664,6 +664,10 @@ class ProcessGroupGloo(ProcessGroupWrapper):
             pg._register_backend(
                 torch.device("cuda"), ProcessGroup.BackendType.GLOO, backend_class
             )
+        if torch.xpu.is_available():
+            pg._register_backend(
+                torch.device("xpu"), ProcessGroup.BackendType.GLOO, backend_class
+            )
         return pg
 
     def getBackendName(self) -> str:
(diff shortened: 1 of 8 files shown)


torchft/process_group.py
d4l3k, line 2084: What's the story for XCCL? Does it support fault tolerance semantics? Do we need BabyXCCL?

d4l3k (approved):
Generally this seems pretty reasonable, will wait on CI

torchft/process_group.py
siju-samuel: Yes. `ProcessGroupXCCL `supports the same fault-tolerance semantics as `ProcessGroupNCCL`, and `ProcessGroupBabyXCCL `is needed for the same reasons `ProcessGroupBabyNCCL `is. 
XCCL follows the same flow as NCCL.

We have future Intel GPUs in the pipeline and want full collective + FT coverage from day zero, so keeping `ProcessGroupXCCL `and `ProcessGroupBabyXCCL `structurally symmetric with the NCCL to avoid future rework.

I've verified `train_diloco.py` and `train_ddp.py `on Intel PVC/BMG devices.

Thanks for the review. The lint issues from the previous CI run are fixed .
Could you please re-approve?

PR #335, opened 2026-07-01
IPv6-safe store address parsing (3 line change)

BenCowen:
## Motivation
Add support for IPv6 fabric.

## Summary

`create_store_client` parses the rendezvous store address with `partition(":")`, taking everything after the **first** colon as the port. That's the correct assumption for `host:port` / IPv4, but it doesn't apply to an IPv6 address, for example on a flat-IPv6 fabric where containers address each other directly by IPv6:

```
store_addr = "fdaa:5137:3ebf:a75:...:29521/torchft/<quorum>/<rank>"
host, _, rest = store_addr.partition(":")     # host = "fdaa"  (!)
port, _, prefix = rest.partition("/")          # port = "5137:3ebf:...:29521"
TCPStore(host_name="fdaa", port=int(port))     # ValueError: invalid literal for int()
```

## Fix

Split off the `/prefix` first, then take the port as the field after the **last** colon, and strip optional brackets from the host. Correct for IPv4, and for bare or bracketed IPv6 (`[host]:port`). One function, no signature change.

## Testing

Verified end-to-end on a flat-IPv6 fabric (Modal i6pn), single-process replica groups across separate containers:
- Lighthouse quorum + cross-replica `allreduce` (Gloo on CPU, and NCCL on GPU).
- DiLoCo training to convergence.
- Fault recovery: a killed replica's replacement heals its checkpoint from a peer over IPv6 (`HTTPTransport`) and rejoins.

Existing `host:port` / IPv4 behavior is unchanged.

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -114,8 +114,12 @@ def create_store_client(store_addr: str, timeout: timedelta) -> Store:
 
     Ex: localhost:1234/my/prefix
     """
-    host, _, rest = store_addr.partition(":")
-    port, _, prefix = rest.partition("/")
+    # IPv6-safe: the address is host:port/prefix (optionally [host]:port/...).
+    # Split off the prefix, then take the port as the field after the LAST colon
+    # so a bare IPv6 address (e.g. a flat-IPv6 fabric) is not mangled.
+    addr_part, _, prefix = store_addr.partition("/")
+    host, _, port = addr_part.rpartition(":")
+    host = host.strip("[]")
 
     store = TCPStore(
         host_name=host,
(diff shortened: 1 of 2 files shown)


BenCowen:
Friendly ping @d4l3k @tushar00jain : 3-line, self-contained IPv6 parse fix, plus a unit test.

d4l3k (approved):
LGTM

d4l3k:
@BenCowen can you rebase on main and fix the lint issues? trunk is green now

I tried to fix it but I don't have permissions to your branch

BenCowen:
Thanks @d4l3k, all set!

PR #336, opened 2026-07-02
Enable Pyrefly for fbcode/torchft (#336)

QuantumManiac:
Summary:

Migrates `fbcode/torchft` from Pyre to Pyrefly by removing the root `python.set_pyrefly(False)` opt-out, so the tree inherits the fbcode root default (`True`). Adds targeted `# pyrefly: ignore` suppressions for pre-existing type errors (64 across 17 files). No runtime behavior changes are intended.

Note: `torchft` is open-source (mirrored to GitHub), so these annotation-only suppressions will need to flow through the OSS export.

Reviewed By: maggiemoss

Differential Revision: D110367354

### DIFF
--- a/torchft/_test/diloco_trainer.py
+++ b/torchft/_test/diloco_trainer.py
@@ -147,6 +147,7 @@ class DiLoCoTrainer:
                     layers.parameters(), lr=0.7, momentum=0.9, nesterov=True
                 )
             )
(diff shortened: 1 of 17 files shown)


maggiemoss (approved):
Review automatically exported from Phabricator review in Meta.

PR #344, opened 2026-08-19
Use the current accelerator for streams and process-group lifecycle

### LINKED ISSUE #340: [RFC] Use the current accelerator for streams and process-group lifecycle
# RFC: Accelerator-neutral streams and process-group lifecycle

## Summary

Use PyTorch's public `torch.accelerator` discovery APIs for the TorchFT operations that own accelerator stream state and process-group lifecycle dispatch. This covers stream contexts, event recording, synchronization, wrapper backend registration, and best-effort lookup of an abort-capable backend.

The proposal does not import a vendor extension, add a backend-specific branch, create a vendor process group, or add a package dependency.

## Problem

TorchFT currently selects CUDA or XPU explicitly in two related infrastructure paths:

1. `torchft/utils.py` chooses the module used for stream contexts, events, and synchronization.
2. `torchft/process_group.py` chooses devices for wrapper backend registration and for fallback backend lookup during abort.

A backend that is otherwise registered through PyTorch's accelerator APIs cannot use these paths without modifying TorchFT. Maintaining separate allowlists also makes the two lifecycle paths drift as new accelerator types are added.

### Stream context

Resolve the current accelerator module dynamically and return its stream context. Keep `nullcontext` for CPU-only execution, a missing stream, or a module without a stream API.

```python
if stream is None or not torch.accelerator.is_available():
    return nullcontext()

accelerator = torch.accelerator.current_accelerator().type
accelerator_module = getattr(torch, accelerator, None)
if accelerator_module is None or not hasattr(accelerator_module, "stream"):
    return nullcontext()
return accelerator_module.stream(stream)
```

### Event recording and synchronization

Create and record an event through the current accelerator module. CUDA retains `interprocess=True`; other backends use their default event constructor. Synchronization targets the current accelerator stream.

```python
accelerator = torch.accelerator.current_accelerator().type
accelerator_module = getattr(torch, accelerator)
event = (
    accelerator_module.Event(interprocess=True)
    if accelerator == "cuda"
    else accelerator_module.Event()
)
accelerator_module.current_stream().record_event(event)
```

```python
if torch.accelerator.is_available():
    accelerator = torch.accelerator.current_accelerator().type
    getattr(torch, accelerator).current_stream().synchronize()
```

### Process-group backend registration

Keep CPU registration and append the current accelerator type when available. Deduplicate CPU-like accelerator reports.

```python
devices = ["cpu"]
if torch.accelerator.is_available():
    accelerator = torch.accelerator.current_accelerator().type
    if accelerator not in devices:
        devices.append(accelerator)
dist.Backend.register_backend(group_name, create_pg, devices=devices)
```

### Abort fallback lookup

A process group with a public `abort()` remains the first choice. Otherwise, resolve the backend for the current accelerator and call its abort method when available. Backend lookup remains best-effort and non-fatal.

```python
backend = None
try:
    if torch.accelerator.is_available():
        backend = pg._get_backend(torch.accelerator.current_accelerator())
except RuntimeError:
    backend = None
if backend is not None and hasattr(backend, "abort"):
    backend.abort()
```

## Ownership Boundary

TorchFT owns generic accelerator discovery and the lifecycle of its wrapper process group. A downstream integration still owns concrete process-group construction when that code depends on backend communication APIs and release-specific runtime behavior.

This change also does not guarantee in-process rejoin after an active collective hangs. Recovery still requires the concrete backend to provide an effective abort or cancellation mechanism.

## Compatibility

- CPU remains registered unconditionally and CPU-only execution remains a no-op or `nullcontext` where appropriate.
- CUDA behavior is preserved, including interprocess event construction.
- XPU uses the same module APIs, selected through `torch.accelerator`.
- Other backends work when the installed PyTorch stack registers the accelerator and exposes the required stream, event, distributed, and optional abort APIs.
- Existing process-group creation, Lighthouse membership, and quorum protocols are unchanged.
- `get_stream_context` uses a backend-neutral return annotation; no runtime call signature changes.

## Non-goals

- Importing or version-pinning a vendor accelerator extension.
- Implementing a backend-specific process group in TorchFT.
- Calling private communication runtime APIs.
- Guaranteeing recovery when the backend cannot cancel a hung communicator.
- Changing LocalSGD, checkpoint, or quorum policy.

## Validation

Ten backend-independent tests cover the combined change:

```text
python -m pytest -q torchft/utils_test.py torchft/process_group_test.py -k <combined-targets>
10 passed
```

They verify CPU/no-accelerator fallback, generic stream and event dispatch, CUDA event compatibility, synchronization, accelerator backend registration, CPU deduplication, current-device backend lookup during abort, and non-fatal lookup failure. They use mocks and require no specialized accelerator hardware.

## Alternatives

Adding another explicit backend branch would solve one integration but keep TorchFT responsible for a growing allowlist. A callback registry would be more flexible, but it is unnecessary for these operations while the public `torch.accelerator` API provides discovery. Concrete process-group construction remains outside this generic dispatch change.

## Rollout

Existing CUDA, XPU, and CPU users require no configuration changes. Integrations for other backends can remove local copies of the generic dispatch logic after adopting the merged implementation while retaining their backend-specific process-group factory.

Skuld7451:
## Summary

- make TorchFT stream, event, and synchronization helpers use the current PyTorch accelerator
- use the current accelerator when registering reconfigurable process groups and resolving an abort-capable backend
- preserve CPU/Gloo and existing CUDA/XPU behavior without adding a backend-specific dependency

Closes #340

## Testing

```text
python -m pytest -q \
  torchft/utils_test.py \
  torchft/process_group_test.py \
  -k "stream_context_without_accelerator_returns_nullcontext or \
      stream_context_uses_current_accelerator_module or \
      record_event_uses_current_accelerator_module or \
      record_event_keeps_cuda_interprocess_event or \
      synchronize_uses_current_accelerator_stream or \
      register_includes_current_accelerator or \
      register_does_not_duplicate_cpu or \
      register_without_accelerator_uses_cpu_only or \
      wrapper_abort_uses_current_accelerator_backend or \
      wrapper_abort_backend_lookup_failure_is_nonfatal"
10 passed
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -331,10 +331,10 @@ class ProcessGroup(BaseProcessGroup):
             return self
 
         devices = ["cpu"]
-        if torch.cuda.is_available():
-            devices.append("cuda")
-        elif torch.xpu.is_available():
-            devices.append("xpu")
+        if torch.accelerator.is_available():
+            accelerator = torch.accelerator.current_accelerator().type
+            if accelerator not in devices:
+                devices.append(accelerator)
         dist.Backend.register_backend(group_name, create_pg, devices=devices)
 
         return group_name
@@ -495,10 +495,10 @@ class ProcessGroupWrapper(ProcessGroup):
             else:
                 backend = None
                 try:
-                    if torch.cuda.is_available():
-                        backend = pg._get_backend(torch.device("cuda"))
-                    elif torch.xpu.is_available():
-                        backend = pg._get_backend(torch.device("xpu"))
+                    if torch.accelerator.is_available():
+                        backend = pg._get_backend(
+                            torch.accelerator.current_accelerator()
+                        )
                 except RuntimeError:
                     backend = None
                 if backend is not None and hasattr(backend, "abort"):
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -9,6 +9,7 @@ import os
 import sys
 from concurrent.futures import Future, ProcessPoolExecutor, ThreadPoolExecutor
 from datetime import timedelta
+from types import SimpleNamespace
 from typing import Any, Callable, cast, Dict, List
 from unittest import skipIf, skipUnless, TestCase
 from unittest.mock import Mock, patch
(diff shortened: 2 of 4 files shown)


d4l3k (approved):
LGTM

d4l3k:
@Skuld7451 can you sign the CLA?

PR #345, opened 2026-08-19
Use the current accelerator for process-group lifecycle

### LINKED ISSUE #341: [RFC] Use the current accelerator for process-group lifecycle
# RFC 0002: Accelerator-neutral TorchFT process-group integration

## Summary

Use the current PyTorch accelerator when TorchFT registers its wrapper backend
and when it asks an underlying process group for a backend-specific abort
implementation. This removes the CUDA/XPU device allowlist without adding
vendor-specific process-group construction code or dependencies to TorchFT.

## Dependency Boundary

This change depends only on PyTorch's current-accelerator discovery API. It
does not import or call the generalized stream helpers, and it does not depend
on FSDP-safe quorum recovery or LocalSGD CPU staging. It can therefore be
adopted directly on the common TorchFT baseline with no merge-order requirement
relative to those capabilities.

An integration still needs to supply a concrete process-group factory and any
backend-specific abort behavior required at runtime. That is an integration
prerequisite, not a source-level dependency between these TorchFT changes.

## Problem

TorchFT's wrapper process group is registered for `cpu` plus an explicitly
detected CUDA or XPU device. On other backends, DeviceMesh and functional
collectives cannot associate accelerator tensors with the wrapper backend. The
abort fallback has the same allowlist and therefore cannot discover an
abort-capable backend for another accelerator.

### Backend registration

At `torchft/process_group.py:321-340`, keep CPU registration and append the
current accelerator type when available.

```python
devices = ["cpu"]
if torch.accelerator.is_available():
    accelerator = torch.accelerator.current_accelerator().type
    if accelerator not in devices:
        devices.append(accelerator)
dist.Backend.register_backend(group_name, create_pg, devices=devices)
```

The duplicate check matters for environments where the reported accelerator is
CPU-like. The backend factory itself remains unchanged and returns the existing
TorchFT wrapper.

### Abort fallback lookup

At `torchft/process_group.py:479-507`, query the underlying process group using
the current accelerator device rather than hard-coded CUDA/XPU devices.

```python
backend = None
try:
    if torch.accelerator.is_available():
        backend = pg._get_backend(torch.accelerator.current_accelerator())
except RuntimeError:
    backend = None
if backend is not None and hasattr(backend, "abort"):
    backend.abort()
```

The fallback remains best-effort. A process group with its own public `abort()`
still takes the first path. Failure to resolve a backend preserves the prior
cleanup behavior of dropping the wrapper reference.

## Ownership Boundary

This RFC deliberately does not create a vendor-specific process group in
TorchFT. The downstream integration owns its process-group factory because it
depends on vendor communication APIs and validated runtime behavior. TorchFT
only needs to avoid excluding a PyTorch-registered accelerator.

Likewise, this change does not promise live rejoin after a backend collective
has already hung. Successful recovery requires the concrete backend to expose a
working abort/cancel mechanism and to tear down the old communicator. Validation
on another backend shows that a surviving replica can remain blocked in an old
all-reduce until the watchdog terminates it; generic Python dispatch cannot fix
that backend limitation.

## Compatibility

- CUDA and XPU are still registered through their current accelerator type.
- CPU remains registered unconditionally.
- No vendor-extension import or install dependency is introduced.
- Backends without `abort` retain prior best-effort cleanup semantics.
- Existing process-group creation and reconfiguration protocols are unchanged.

## Non-goals

- Adding a vendor-specific process-group implementation to TorchFT.
- Calling vendor-private communication APIs.
- Guaranteeing in-process rejoin after an active collective failure.
- Changing Lighthouse membership or quorum policy.

## Related Work

Issue `#257` and PR `#307` cover broader device-agnostic initialization and
concrete accelerator process-group wrappers. This RFC is narrower: it removes
the device allowlist when the existing TorchFT wrapper registers its backend
and when it discovers an abort-capable underlying backend. It neither replaces
those concrete wrappers nor changes how an integration constructs them.

## Validation

Five backend-independent unit tests cover the changed branches:

```text
python -m unittest -v \
  torchft.process_group_test.ProcessGroupTest.test_register_includes_current_accelerator \
  torchft.process_group_test.ProcessGroupTest.test_register_does_not_duplicate_cpu \
  torchft.process_group_test.ProcessGroupTest.test_register_without_accelerator_uses_cpu_only \
  torchft.process_group_test.ProcessGroupTest.test_wrapper_abort_uses_current_accelerator_backend \
  torchft.process_group_test.ProcessGroupTest.test_wrapper_abort_backend_lookup_failure_is_nonfatal
Ran 5 tests
OK
```

They verify accelerator registration, CPU deduplication, the CPU-only fallback,
current-device backend lookup during abort, and non-fatal behavior when backend
lookup fails. All tests use mocks and require no specialized hardware.

## Alternatives

Placing all vendor-specific process-group creation and registration in TorchFT
would couple the framework to vendor release details. Keeping all logic in a
downstream integration would avoid a TorchFT change, but TorchFT's fixed
CUDA/XPU allowlist would remain an artificial blocker for every non-listed
accelerator. The proposed boundary is the smallest backend-neutral change.

## Rollout

After this change is merged into TorchFT, downstream integrations can continue
supplying their process-group factories. They only remove duplicated accelerator
allowlist logic, not vendor-specific process-group implementations.

Skuld7451:
## Summary

- use the current accelerator when registering reconfigurable process groups
- use the selected accelerator when aborting process groups
- preserve the existing CPU/Gloo path

Closes #341

## Testing

```text
python -m pytest -q \
  torchft/process_group_test.py \
  torchft/manager_test.py
5 targeted tests passed
```

### DIFF
--- a/torchft/process_group.py
+++ b/torchft/process_group.py
@@ -331,10 +331,10 @@ class ProcessGroup(BaseProcessGroup):
             return self
 
         devices = ["cpu"]
-        if torch.cuda.is_available():
-            devices.append("cuda")
-        elif torch.xpu.is_available():
-            devices.append("xpu")
+        if torch.accelerator.is_available():
+            accelerator = torch.accelerator.current_accelerator().type
+            if accelerator not in devices:
+                devices.append(accelerator)
         dist.Backend.register_backend(group_name, create_pg, devices=devices)
 
         return group_name
@@ -495,10 +495,10 @@ class ProcessGroupWrapper(ProcessGroup):
             else:
                 backend = None
                 try:
-                    if torch.cuda.is_available():
-                        backend = pg._get_backend(torch.device("cuda"))
-                    elif torch.xpu.is_available():
-                        backend = pg._get_backend(torch.device("xpu"))
+                    if torch.accelerator.is_available():
+                        backend = pg._get_backend(
+                            torch.accelerator.current_accelerator()
+                        )
                 except RuntimeError:
                     backend = None
                 if backend is not None and hasattr(backend, "abort"):
--- a/torchft/process_group_test.py
+++ b/torchft/process_group_test.py
@@ -9,6 +9,7 @@ import os
 import sys
 from concurrent.futures import Future, ProcessPoolExecutor, ThreadPoolExecutor
 from datetime import timedelta
+from types import SimpleNamespace
 from typing import Any, Callable, cast, Dict, List
 from unittest import skipIf, skipUnless, TestCase
 from unittest.mock import Mock, patch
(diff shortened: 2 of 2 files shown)


Skuld7451:
This change is now included in #344 so accelerator stream utilities and process-group lifecycle handling can be reviewed together. Closing this superseded PR; its source branch is retained for history.

PR #355, opened 2026-09-04
Fix managed-future callback late binding and record_event return value

romanushakov-synth:
## What is broken

This PR fixes two bugs. Both break training when we use `ProcessGroupBabyNCCL` (the process group that runs NCCL in a child process).

**Bug 1: the callback chain reads the wrong node.**
`_ManagedWork` keeps a list of callbacks and hooks them up later, in a loop. Each callback was supposed to remember its own spot in the list. But because of how Python closures work, all callbacks shared one variable, and the loop kept moving that variable forward. When the loop was done, the variable pointed at the end of the list, which has no callback (`None`).

If the result is ready right away (like in the existing tests with `_DummyWork`), the callback runs inside the loop and sees the right spot. So nothing looked wrong. But with `ProcessGroupBabyNCCL`, the result comes later, from another thread. Then every callback saw the end of the list, hit `assert None`, and failed. So every allreduce failed and no step could finish.

**Bug 2: `record_event` threw the event away.**
`record_event()` made a CUDA event but returned nothing. All callers do `event = record_event()` and send the event to another process, which calls `event.wait()` before reading the tensor. Since the event was always `None`, the wait was skipped. The reader could read the tensor before the GPU finished writing it. This fails silently and only sometimes, when the collective is slow.

## The fix

- In `_set_future_callback`, give each callback its own copy of its list node (`node=managed_fut` as a default argument), instead of the shared variable.
- Make `record_event()` return the event it records.

## How I tested it

- New tests in `manager_test.py` build the callback chain first and complete the future from another thread, like baby NCCL does. They fail on `main` and pass with this fix.
- New test in `utils_test.py` checks `record_event()` returns a real, shareable CUDA event. It fails on `main` and passes with this fix.
- `manager_test.py` + `utils_test.py`: 26/26 pass.
- `process_group_test.py` gives the same result on this branch and on `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/torchft/manager.py
+++ b/torchft/manager.py
@@ -1280,17 +1280,21 @@ class _ManagedWork(dist._Work):
         is_future_wrapped = False
         while managed_fut._next:
 
+            # Bind the node as a default argument: the loop rebinds `managed_fut`,
+            # so a callback that runs after the loop (e.g. when the future completes
+            # on another thread) would otherwise see the tail node instead.
             def callback(
                 fut: torch.futures.Future[object],
+                node: _ManagedFuture[object] = managed_fut,
             ) -> object:
-                nonlocal managed_fut, value
+                nonlocal value
                 # change the stream to avoid making the callback stream
                 # dependent on process group stream running the allreduce
                 with get_stream_context(self._stream):
                     # Setup stream dependency
                     fut.wait()
-                    assert managed_fut._callback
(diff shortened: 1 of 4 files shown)


d4l3k (approved):
LGTM

PR #356, opened 2026-09-20
Use PGTransport constructor timeout as the transfer default

### LINKED ISSUE #323: `PGTransport.__init__(timeout=...)` is effectively ignored by checkpoint send/recv
### Summary

`PGTransport` stores `timeout` in `__init__` (`self._timeout`), but `send_checkpoint` / `recv_checkpoint` use the method argument `timeout` for all `work.wait(...)` calls.

In practice, transfer timeout is controlled by the caller (typically `Manager._timeout`), not by `PGTransport(timeout=...)`.

This creates ambiguous API semantics: constructor-level timeout appears configurable but has no effect for checkpoint transfer.

### Expected behavior

If `PGTransport` accepts a constructor-level `timeout`, it should either:

- be used as a default timeout in checkpoint transfer paths, or
- be clearly documented as unused/deprecated.

### Actual behavior

Constructor timeout is stored but not used in checkpoint transfer wait paths.

### Repro

1. Construct `PGTransport(pg, timeout=timedelta(seconds=10), ...)`.
2. Call `send_checkpoint(..., timeout=timedelta(seconds=60))` or `recv_checkpoint(..., timeout=timedelta(seconds=60))`.
3. Observe `work.wait(...)` always uses the per-call `timeout` argument, not `self._timeout`.

### Suggested fix

Minimal and explicit options:

1. Remove `timeout` from `PGTransport.__init__` and keep timeout explicit per call.
2. Keep both and define/document precedence (e.g. per-call `timeout` fallback to `self._timeout`).

### Optional follow-up

Split `Manager.timeout` into control-path timeout and `checkpoint_timeout`, since checkpoint transfer latency is often very different from fast RPC/control calls.

rohan-patnaik:
Fixes #323.

Use the `PGTransport` constructor timeout when checkpoint send/receive omits `timeout` or passes `None`. Explicit per-call values, including zero, stay unchanged; Manager still supplies its own timeout. Document the per-wait contract and add real CPU/Gloo regression tests.

Validation: Linux checkpoint suite: 23 passed, 3 GPU skips; eight added cases fail on base. Pyre and formatting pass. Independent Mac review: all 14 added cases pass in a scratch copy with Darwin skips removed; changed-file Flake8 and the docs build with `-W` pass. Four existing Mac HTTP test failures repeat on base.

AI assistance was used for implementation and independent review (Claude Fable 5.1, High).


d4l3k (approved):
LGTM

